PR #4, opened 2019-09-02
HTTPS link to twitter and remove www

coliff:
saves a redirect


tatianamac:
Ah, really good call. Thank you so much, @coliff 💙

PR #10, opened 2019-10-20
Remove os fluff files

lynncyrin:
## Changes

- removes `.DS_Store` files
- adds `.gitignore` for `.DS_Store` files

## About `.DS_store`

This type of file is commonly removed and ignored from git repos, here's a fairly popular stackoverflow question about it https://stackoverflow.com/questions/107701/how-can-i-remove-ds-store-files-from-a-git-repository

## About `.gitignore`

docs: https://git-scm.com/docs/gitignore

## [meta] About this PR description

I'm hoping to provide enough context here for any random person who might drop into this PR - rather making assumptions about what context the maintainers (eg. Tatiana ^^) would need.


tatianamac:
Eeek! Thank you, @lynncyrin ! 💙

PR #28, opened 2019-11-08
➕ Improve Fatphobia definition and update further reading links

monika:
I updated the `Further Reading` section with a more comprehensive resource page. I also filled out the `Issues` and `Impact` section, but I'm no writer so I'd appreciate any feedback on my copy 🙏🏻

### DIFF
--- a/_site/index.html
+++ b/_site/index.html
@@ -514,7 +514,7 @@
             <p class="word__type">
               Issues
             </p>
-            <p class="word__breakdown">Fatphobia is the racialized, classist, and ableist concept of an &ldquo;ideal&rdquo; body type that centers a white, Western aesthetic ideal and endeavours to form a direct correlation between thinness and heathliness&mdash;perpetuating white supremact violence against racialized people, and the ableist idea of healthiness as an indicator of human value.</p>
+            <p class="word__breakdown">Fatphobia is the racialized, classist, and ableist tool of oppression that codifies an &ldquo;ideal&rdquo; body type that centers a white, Western aesthetic ideal, demands time and resources that people may not have, and attempts to correlate thinness and heathliness&mdash;perpetuating white supremact violence against racialized people, classist oppression of people living with limited resources, and the ableist idea of healthiness as an indicator of human value.</p>
           </div>
           <div class="block__type">
             <p class="word__type">
--- a/_site/index.html
+++ b/_site/index.html
@@ -522,7 +522,7 @@
             </p>
             <p class="word__breakdown">
               Medical professionals hold fatphobic biases and often prescribe weight loss instead of listening to their fat patients' concerns. That history of being ignored and shamed keeps many fat people from visiting medical professionals even if they have the means.<br><br>
-              Employeers hold fatphobic biases and percieve their fat employees as lazy, less intelligent, and unmotivated, therby paying fat employees less than their thin co-workers.<br><br>
+              Employeers hold fatphobic biases and percieve their fat employees as lazy, less intelligent, and unmotivated, therby paying fat employees less than their thin co-workers and offering them no protection from weight stigma in the workplace.<br><br>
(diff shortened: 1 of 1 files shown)


_site/index.html
```diff
@@ -510,6 +510,20 @@ <h3>T-Z</h3>
           <p class="word__definition">
             the fear, hatred, and stigmitazation of fat people.<br><br>
           </p>
+          <div class="block__type">
+            <p class="word__type">
+              Issues
+            </p>
+            <p class="word__breakdown">It perpetuates the idea that there is a “normal” body type, that there is a direct correlation between thinness and heathliness (which also perpetuates healthiness as an indicator of value), and that fatness is both
```
tatianamac: Do you feel comfortable adding factors around class and race?

Typically fatphobia exacerbates racialisation of what an "ideal" body type is as well through a white and Western lens.

_site/index.html
```diff
@@ -510,6 +510,20 @@ <h3>T-Z</h3>
...
+            abnormal and unhealthy.</p>
+          </div>
+          <div class="block__type">
+            <p class="word__type">
+              Impact
+            </p>
+            <p class="word__breakdown">Fat people have a difficult time accessing medical care, they earn less money, are less likely to be believed when reporting sexual assaults, and have
```
tatianamac: I understand what you mean here, but I think that more connecting of the dots would be helpful through rephrasing, such as "More hurdles are placed before fat people such as X, which prevent them from accessing medical care, earning less money, etc." 

I want to ensure the systems of oppression that prevent them are explicitly stated.

_site/index.html
```diff
@@ -518,7 +532,7 @@ <h3>T-Z</h3>
               <span class="word__alt">
                 <a class="word__link"
                   href="https://www.srhweek.ca/providers/people-and-communities/sizeism-and-fatphobia/">Sizeism and Fatphobia</a>;
-                <a class="word__link" href="https://www.huffpost.com/entry/fatphobia-a-guide-for-the_b_5214754">Fatphobia: A Guide for the Disbeliever</a>.
+                <a class="word__link" href="https://simmons.libguides.com/anti-oppression/anti-fatmisia">Anti-Oppression: Anti-Fatmisia</a>.
```
tatianamac: I'm curious if you think we should to add an `Alt Words` section here to help illuminate the idea of "-phobia" versus "anti-" or "-misia" as we discussed.

_site/index.html
tatianamac: Suggest writing this as "Fatphobia perpetuates..." so that we don't start with the empty word "It."

tatianamac:
@monika : Do you have time/energy to update this definition? I'm eager to get this word included in the dictionar, and also want to be mindful of your time. Thank you for all you've done already!

monika:
@tatianamac Sorry about the wait, with the holidays I haven't had as much personal time as I'd hoped but I think I can get these edits completed before the weekend is over. Thanks much for taking the time to review everything and provide feedback 🙌

_site/index.html
monika: Yes, absolutely - those are important factors, thank you for catching that.

_site/index.html
monika: I think that would work, for sure, although "fatphobia" itself is an important word to know because the community uses it, plus we have so much literature that uses "fatphobia".

What do you think about it being more of a generalized entry that describes the nuance between those suffixes/prefixes, too, because it can apply to other entries within the dictionary?

_site/index.html
monika: I've updated the `Impact` section to outline the specific facets of fatphobia used by the three different institutions I brought up and how they use them against fat people.

monika:
@tatianamac I've added some changes that I think tackle the updates you requested. Let me know what you think and if there's anything else you'd like to see!

PR #30, opened 2019-11-08
➕ Propose definition of polyamory

avo7:
More explanation: https://avris.it/blog/my-definition-of-polyamory

### DIFF
--- a/_site/index.html
+++ b/_site/index.html
@@ -764,6 +764,15 @@
               </span>
             </p>
           </div>
+
+          <div class="block__type">
+            <p class="word__type">
+              Further Reading
+            </p>
+            <ul class="word__breakdown">
+              <li><a class="word__link" href="https://avris.it/blog/my-definition-of-polyamory" rel="noreferrer">My definition of polyamory</a></li>
+            </ul>
+          </div>
         </section>
 
         <section class="block__word list">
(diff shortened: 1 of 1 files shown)


_site/index.html
```diff
@@ -739,6 +739,33 @@ <h3>T-Z</h3>
...
+
+          <div class="block__type">
+            <p class="word__type">
+              Note
+            </p>
+            <p class="word__breakdown">
+              <span class="word__alt">
+                One can be polyamorous despite currently being single or in a monogamous relationship.
+              </span>
+            </p>
```
tatianamac: Is there a reason you excluded the resource you shared in the comment from a `Further Reading` section? If not, I think any links are always helpful for people to read more.

avo7: Because I only thought about writing this blog post a day after I have submitted the PR 😅
If you think it's the right place for this link, I'll add it :)

tatianamac: 🙈😅 That makes sense! I'd love to include your blog post (as I thought it was very personal and helpful to me, and I appreciated the shoutout to the dictionary).

Maybe you could add 1 or 2 external ones that you like to help round out the perspective? (I try to include a few per word to illustrate the richness of words.) I trust your suggestions.

PR #43, opened 2020-01-15
👩‍🎨 Avoid Breaks in Table of Content Items

ovlb:
Due to auto-flowing of CSS columns browsers were free to break content inside of a ToC item, or list of subterms. This lead to subterms at the beginning of a column. 

E.g. here is Chrome breaking in the subterms of black:

Snapshot of the multi-column view on the homepage, where the list of subterms is broken in a way that they are disconnected from the main term.

By using `break-inside: avoid` ([MDN](http://developer.mozilla.org/en-US/docs/Web/CSS/break-inside)) this problem gets fixed.

I also started to create module files for CSS to make it easier to find elements.

### DIFF
--- a/11ty/_includes/components/table-of-content-item.njk
+++ b/11ty/_includes/components/table-of-content-item.njk
@@ -1,4 +1,4 @@
-<li>
+<li class="toc__list-item">
   {{ definition.data.title | linkIfExistsInCollection(collections.definedWords) | safe  }}
   {%- if
     definition.data.flag and
--- a/11ty/_includes/components/table-of-content.njk
+++ b/11ty/_includes/components/table-of-content.njk
@@ -1,16 +1,16 @@
 <section>
   <nav class="" aria-label="Definitions">
     <ol class="multi-column u-no-padding-left list">
-    {% for section in collections.tableOfContent %}
-      <li>
-        <span class="sub-headline">{{ section.title }}</span>
-        <ol>
-          {% for definition in section.definitions %}
-            {% include 'components/table-of-content-item.njk' %}
-          {% endfor %}
-        </ol>
-      </li>
-    {% endfor %}
+      {% for section in collections.tableOfContent %}
+        <li>
+          <span class="sub-headline">{{ section.title }}</span>
+          <ol class="toc__list">
+            {% for definition in section.definitions %}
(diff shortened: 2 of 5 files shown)


tatianamac:
@ovlb I missed this when it came through—thank you for your patience!

All of the CSS updates look good.

I have a separate concern: I'm wondering if it's intentional that we have .scss files in the css folder; I typically keep those separate and have scss folder compile into a css folder, such that I can have sass watch the scss folder and compile to the css folder. I could be missing something here!

ovlb:
@tatianamac Yip. I would prefer this, too. I would prefer to open a separate clean-up PR also including the removal of the old `_site` folder mentioned in https://github.com/tatianamac/selfdefined/pull/50#issuecomment-582789609. Is this alright with you?

dengeist:
@olvb, @tatianamac I actually have some ideas about the scss cleanup. while investigating how I'd like to build my own site, I discovered an [Eleventy sass plugin](https://github.com/Sonaryr/eleventy-plugin-sass). It could be used in selfdefined to completely remove the need for Parcel!

ovlb:
@dengeist Interesting! Have you tried it already? Do you know if it can be integrated with PostCSS? Adding it one thing I had in mind to add later, to add backwards compatibility for things like custom properties.

Also, Parcel will come in handy once we start building more client-side functionality (like collecting definitions for your own page or something).

dengeist:
@ovlb I have tried it! Thanks for reminding me that Parcel will have other uses for the app. I would be happy to start a discussion about this as an RFC if you would like!

ovlb:
@dengeist Thanks for the offer! I will gladly take you up on it, once this gets closer to reality :)

olvb:
Hi all,
I am @olvb on github and i am receiving messages intended for @ovlb.

dengeist:
Very sorry about that, @olvb! Thank you for bringing this up. I’ll be more careful with my typing. 😅

tatianamac:
Closing out this issue as I believe it's resolved.

@dengeist Please feel free to open the RFC for now an we can label it accordingly (so that as @ovlb points out, when the time comes, we can reinvigorate discussion around it) and migrate this discussion there.

PR #44, opened 2020-01-17
Add definition for the word barbarian

codeAbul:
- Adds definition for the word barbarian including the impact, usage tip, issues and sources. 

( Side note: This is my first ever contribution to a OSS project. Thanks to @tatianamac for being so welcoming, giving the confidence that OSS is not the domain of cishet white bois alone )

PR #45, opened 2020-01-17
🏴 Convert flag level to lowercase before trying to get infos

ovlb:
As experienced in #44 the build breaks if you add a flag and write its level in title case (e.g. `Avoid` instead of `avoid`). 

To make it more resilient I always convert the flag level to lower case whilst getting the info.

The changeset is a bit convoluted because Prettier wasn’t working when I wrote the config but is now. The relevant commit is https://github.com/tatianamac/selfdefined/commit/f12c5c9fedd99b68143a45436a6692934f95e5df

### DIFF
--- a/.eleventy.js
+++ b/.eleventy.js
@@ -66,7 +66,7 @@ module.exports = function(config) {
     ]);
 
     if (flag) {
-      const info = cleanText.get(flag.level);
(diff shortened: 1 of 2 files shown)


tatianamac:
🚀

PR #65, opened 2020-02-10
Set $primary-color as focus indicator

### LINKED ISSUE #59: Proposal: use a custom focus indicator
## Problem
The default focus indicators provided by some browsers and operating systems are inaccessible to many users. The below screenshot of selfdefined illustrates the default focus indicator in Firefox: a thin, dotted outline.

## Solution
Use the `:focus` pseudoclass to define a custom color and style for the focus outline in selfdefined. Ensure that the color contrast is at least 3:1 with the background of the app.

The [`$primary-color` defined in `base.css`](https://github.com/tatianamac/selfdefined/blob/master/assets/css/base.scss#L34) would be a great solution here!
## Further reading
- [Understanding Success Criterion 2.4.7: Focus Visible](https://www.w3.org/WAI/WCAG21/Understanding/focus-visible.html)
- [Understanding Success Criterion 1.4.11: Non-text Contrast](https://www.w3.org/WAI/WCAG21/Understanding/non-text-contrast.html)

xavierbellhaddon:
Fixes #59.

I also added a 3px outline offset. Let me know if you'd like me to change that!

### DIFF
--- a/assets/css/base.scss
+++ b/assets/css/base.scss
@@ -114,6 +114,10 @@ a {
   &:hover {
     border-bottom: red solid 0.1rem;
   }
+  &:focus {
+    outline: 3px solid $primary-color;
+    outline-offset: 5px;
+  }
 }
 
 .subtitle {


tatianamac:
@ovlb I will open a separate issue for this, but for context for use cases like this I think it makes sense to write a Sass function of sorts that will convert 5px to rem

PR #69, opened 2020-02-12
Add a Sass function to convert pixels to rems (resolves #67)

### LINKED ISSUE #67: ⚒️ Add Sass function to convert px to rem
Add a Sass function to convert `px` to `rem`, so that contributers can still reference/use `px` but with the responsiveness of `rems`

Guide article:
https://css-tricks.com/snippets/sass/px-to-em-functions/

greatislander:
This pull request adds a function for converting pixels to rems as requested in #67. It uses the unitless example from [CSS Tricks](https://css-tricks.com/snippets/sass/px-to-em-functions/), so the following will both work:

```scss
body {
  font-size: rem(20);
}
```

```scss
body {
  font-size: rem(20px);
}

### DIFF
--- a/assets/css/base.scss
+++ b/assets/css/base.scss
@@ -3,13 +3,29 @@
 
 @import '~prism-themes/themes/prism-a11y-dark';
 
+// FUNCTIONS
+
+$browser-context: 16;
+
+@function pxToRem($pixels, $context: $browser-context) {
+  @if (unitless($pixels)) {
+    $pixels: $pixels * 1px;
+  }
+
+  @if (unitless($context)) {
+    $context: $context * 1px;
+  }
+
+  @return $pixels / $context * 1rem;
+}
+
 pre[class*='language-'] {
   font-size: 1rem;
 }
 
 code {
   background-color: #333333;
-  border: 1px solid black;
+  border: pxToRem(1) solid black;
   color: white;
   padding: 0.15em;
   border-radius: 0.25em;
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by tatianamac** [comment]
Thank you, @greatislander! This is so awesome and helpful.

@ovlb would love a second review here. I think we'll also want to open up an issue to ensure that we change all instances of px to rems using this function.

**Comment by greatislander** [comment]
@tatianamac Glad to contribute! I’m also happy to take care of the follow-up issue you mentioned, but I figured I would wait to let you review this before jumping ahead to converting everything to use the function :)

File: assets/css/base.scss [hunk 1]
```diff
@@ -77,10 +77,26 @@ $bold: 700;
   margin-right: 0.35rem;
 }
 
+// FUNCTIONS
+
+$browser-context: 16;
+
+@function rem($pixels, $context: $browser-context) {
```
ovlb: The function works fine. Only thing: I think the name should be a tad more concrete. Something like `pxToRem`. To make it clear for other contributors that it won’t convert, say, `ch` or percentage units.

```suggestion
@function pxToRem($pixels, $context: $browser-context) {
```

greatislander:
Sounds good @ovlb!

greatislander:
One other thought occurred to me — structurally, would there be some value in moving Sass functions and mixins into their own partials that are imported at the top of the base file?

ovlb:
> One other thought occurred to me — structurally, would there be some value in moving Sass functions and mixins into their own partials that are imported at the top of the base file?

Yes, there definitely is. I started this work for the regular classes in #75 but skipped the Sass part, as I waited for this PR to get merged. 
Any preference on how to do it? In other projects, I have a folder named `_sass` in which I collect such partials. But am open to suggestions :) If you want to do this, go ahead!

greatislander:
@ovlb I think having `_mixins.scss` and `_functions.scss` partials in that `_sass` folder would make sense. I can move the function into such a partial in this PR — if you prefer I could make a separate one for the mixins so as to keep this PR within scope.

ovlb:
@greatislander That sounds perfect, thanks!

greatislander:
@ovlb Done!

PR #76, opened 2020-02-15
🏡 eleventy@0.10.0

ovlb:
There has been a new version we missed. I updated our package.


ovlb:
🤷‍♀️ 

Will further investigate the build failure later.

ovlb:
@tatianamac I updated again and specified Node 12 as the version it has to run, which seems to have fixed it.

ovlb:
Don’t know why the changelog is so convoluted, though.

PR #78, opened 2020-02-16
🏡 Install ESLint && Ava

ovlb:
Integrate linting and unit testing.


ovlb:
I’ve also sorted the scripts in package.json alphabetically, because because. 🙈

Issue #41, opened 2020-01-15
🎨 Fix red flag splitting across two lines


laurakalbag:
<img width="300" alt="Screenshot 2020-01-15 at 14 47 06" src="https://user-images.githubusercontent.com/266663/72443468-36b2ae80-37a6-11ea-866a-f631eb28ea54.png">

Depending on the column flow, flags are occasionally split across two lines of text. More often it happens with longer words and longer flags.

michaelspellacy:
Hi @tatianamac. Busy week. Will swing back round to this tomorrow.

PR #79, opened 2020-02-18
Fixed red flag splitting across two lines

michaelspellacy:
Addressed Issue #41 Added white-space: nowrap to .flag_red and also added new properties to link within toc to compensate for nowrap. Tested across several browsers and devices. https://www.screencast.com/t/zAm8KP7CPciC Hope this helps! 

p.s. Love what you are doing here!


assets/css/structures/_table-of-content.scss
```diff
@@ -1,3 +1,9 @@
 .toc__list-item {
   break-inside: avoid;
+
+    a {
+    display: inline-block;
+    margin: 0.5rem 0.75rem;
```
tatianamac: This `margin` seems to be adding some extra spacing around each linked word. Is this visual adjustment necessary for the red flags to not break?

If it's not necessarily, I'd prefer the alignment to be left-aligned without the extra padding.

_Left is original where all text is aligned left_; right is PR where there is extra spacing

Side-by-side of [1] correctly rendered text (without a leading space) and [2] incorrectly rendered text (with a leading space)

assets/css/base.scss
```diff
@@ -419,12 +419,13 @@ th {
 
 .flag__red {
   background-color: rgb(255, 192, 203);
+  border-radius: 1rem;
   font-size: 0.9rem;
   font-weight: bold;
-  border-radius: 1rem;
-  padding: 0.45rem 0.65rem;
   margin: 0.25rem 0.75rem;
+  padding: 0.45rem 0.65rem;
```
tatianamac: More a curiosity than an issue if you don't mind explaining (for my own edification): Why did you rearrange these properties? 🤓

assets/css/structures/_table-of-content.scss
```diff
@@ -1,3 +1,9 @@
 .toc__list-item {
   break-inside: avoid;
+
+    a {
+    display: inline-block;
```
tatianamac: The `inline-block` here causes an issue with the underline for words that become multiple lines, like OCD:

Obsessive Compulsive Disorder has underline that extends beyond the word

assets/css/structures/_table-of-content.scss
ovlb: I think it’s fine to have `margin-left: 0` and go with the slightly indented flag. Which happens with non defined words, too.

tatianamac:
@michaelspellacy Are you still able to address these comments? No rush on our end as I imagine you're busy, but want to check in. :)

assets/css/base.scss
michaelspellacy: @tatianamac  Sorry again for delay! Day job got out of hand! I simply reorder them alphabetically out of habit for better readability, but happy to change them back in preference to what you may already have in place throughout CSS.

tatianamac: No sorries. Understood; carry on as you are. I was mostly curious.

michaelspellacy:
Hi @tatianamac and @ovlb!

Took a poke at revisiting this. You'd think it was easy, but no, it never is. :-) This is what I did:

**eleventy.js** - I really felt that a new span was necessary to get us the hook we needed to balance styles on terms with and without hyperlinks. 

**_flag.scss** - Restored old order. Adjusted margin. Added display: inline-block;

**_table-of-contents.scss** - Included <a> and <span> elements (feel free to add a shared class to these in the future). Added margin-right: 0.75rem since it was removed from left margin of .flag__red

Think it looks much better now. Let me know if any new issues!

tatianamac:
@ovlb Since you updated the sass last, do you mind taking a look over this? I'll close it once you give it 🐚 of approval

ovlb:
With thanks to @michaelspellacy my sea( shel)l of approval is given. 🐚

michaelspellacy:
Thanks gang! Glad I could help! ❤️

Issue #97, opened 2020-03-19
Load fonts progressively


dengeist:
Problem
---
Right now, the Typekit font bundle significantly slows page-load times. In limited local testing, it adds >1 second to load time on a simulated "average" 3G connection. This number increases significantly on lower-quality simulated connections, leading to average load times >5 seconds on "regular 2G" simulations.

Since a significant number of users browse the web on their phones, we should do our best to load these fonts progressively, so that they do not block user interaction on the page.

Proposal
--- 
Use Filament Group's [LoadCSS library](https://github.com/filamentgroup/loadCSS) to bring support for link rel="preload" to all browsers. 

With lazy-loading established, we can use it in the future for other non-critical styles that we may write for the app.

Benefits & trade-offs
---
➕ Page load time will decrease; user will be able to read, interact with our content sooner

➖ Browser may re-paint if there's a significant delay in loading the typekit bundle

tatianamac:
Thanks, @dengeist ! Sounds good. 

I'm also wondering if as a multi-fold approach we can also upload/locally load the fonts as well and `font-display: swap` as well.

dengeist:
This crossed my mind as well! All current browsers but IE 11 support the `font-display` property (source: [caniuse](https://caniuse.com/#feat=css-font-rendering-controls)), so we would have pretty good coverage if we moved to local font files. Additionally, hosting the fonts ourselves would cut out any reliance on TypeKit's servers. We might even get a bit of a site performance boost since we'd be minimizing requests to other domains.

Lazy-loading would still come in handy for users on older browsers that do not support `font-display`!

tatianamac:
Perfect! Would you be able to make those adjustments if I got you all the `woffs`? I haven't interacted with Typekit in a moment, but I *believe* that's all possible. Otherwise I will figure something out as a substitute.

Also when you say lazy-loading here do you mean preloading? (Or am I confuse?)

dengeist:
I would be able to implement this if you got all the raw `.woff` files, yeah! If you can't, using `rel="preload"` and ensuring support for it should be a serviceable alternative!

(yes, by "lazy-loading", I do mean "preloading" here. preloading is lazy-loading in the context of CSS files.)

tatianamac:
[...]

PR #98, opened 2020-03-19
Add a `pageStyles` block to base template

dengeist:
Also: 
- Imports the TypeKit font bundle as a link in the base template
- Imports the `prism-a11y-dark` theme as a link in only the documentation template 

Importing these external stylesheets bundle as links will improve UX because css `@import`s are render-blocking, while `<link>` imports are not. Since most users spend time on the main page, moving the prism link to the documentation template will prevent unnecessary downloads.

Importing the TypeKit bundle as a link will also make #97 (Load fonts progressively) possible

### DIFF
--- a/11ty/_includes/layouts/base.njk
+++ b/11ty/_includes/layouts/base.njk
@@ -11,7 +11,7 @@
       name="description"
       content="{{ renderData.description or description or metadata.description }}"
     >
-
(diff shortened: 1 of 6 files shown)


tatianamac:
V. helpful—thank you, @dengeist !

tatianamac:
LOL my chub thumb accidentally hit close and merge. Thanks again!

PR #105, opened 2020-03-31
Import fonts with rel=preload polyfill

### LINKED ISSUE #97: Load fonts progressively

mxmason:
Closes #97 (Load fonts progressively)

This approach directly copies an inline, minified version of the rel="preload" polyfill to the bottom of the `base.njk` template, per Filament Group's suggestion. A less-fragile future approach might be to:
- Add fg-loadcss as a dependency
- copy the contents of the minified js file at build time
- inline the js file into `base.njk` with something like [this filter](https://www.11ty.dev/docs/quicktips/inline-js/)

### DIFF
--- a/11ty/_includes/layouts/base.njk
+++ b/11ty/_includes/layouts/base.njk
@@ -11,7 +11,9 @@
       name="description"
       content="{{ renderData.description or description or metadata.description }}"
     >
-    <link rel="stylesheet" href="https://use.typekit.net/qlo3dpu.css" rel="preload"/>
+
+    <link rel="preload" href="https://use.typekit.net/qlo3dpu.css" as="style" onload="this.onload=null;this.rel='stylesheet'">
+    <noscript><link rel="stylesheet" href="https://use.typekit.net/qlo3dpu.css"></noscript>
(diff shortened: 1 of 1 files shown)


dengeist:
Hi @tatianamac and @ovlb! Please read [this comment on my original issue](https://github.com/tatianamac/selfdefined/issues/97#issuecomment-607658286) before you decide to merge this. Thank you!

PR #106, opened 2020-04-02
Adding link changes with text-decoration and background

### LINKED ISSUE #102: Firefox: Focus outline obscures text when link wraps multiple lines
## Problem
The following screenshots show a link reading "Obsessive Compulsive Disorder (OCD)" as seen on the Self-Defined website, with the site's red custom focus drawn around it.

**In Firefox**, the outline is drawn around _each line_ of the wrapped text, making the text hard to read.
Word_Wrap_Focus_FF

**In Safari and Chrome**, the outline is drawn around the whole anchor element. This way, the outline does not obscure any text. 
Word_Wrap_Focus_Safari

## Possible fixes 
@wuz (Conlin) suggested using text-decoration styling to restore word-level underline to links when they are rendered as `inline-block`s. Here's [Conlin's codepen demo](https://codepen.io/wuz/pen/BaNPbNb).

The [support matrix for text-decoration styling on caniuse](https://caniuse.com/#feat=text-decoration) looks good.

**Edit:** 
I was initially stoked about Conlin's suggestion, but further research into custom underline properties made me sad. :(
We could also:
- display links as inline-block
- add a child span to them
- apply a bottom border to the child span instead of the link

This approach works in more browsers – even modern browsers do not support custom underlines with text-decoration properties – and is most like our current styles. I made [a CodePen to compare the two approaches](https://codepen.io/dengeist/pen/mdJjYmB).

## Details for nerds
As evidenced in [the Firefox source code](https://searchfox.org/mozilla-central/rev/f1f75f0a2d995241fbc454edfab4be1064c544e1/layout/painting/nsDisplayList.cpp#5389 ), this seems to be a known issue related to how Firefox chooses to draw the focus outline.

wuz:
Changes some link styles to fix #102!

Screen Shot 2020-04-01 at 11 54 16 PM


assets/css/base/_a.scss
```diff
@@ -1,16 +1,17 @@
...
   &:focus {
-    border-bottom: $primary-color solid 0.1rem;
+    text-decoration-color: $primary-color;
   }
   &:focus {
-    outline: pxToRem(3) solid $primary-color;
-    outline-offset: pxToRem(5);
+    background-color: pink;
+    color: $black;
   }
```
dengeist, line 16:
```suggestion
    outline: 0;
  }
```

@wuz I suggest we add in `outline:0` here. If we don't, the browser-default focus ring will still show, and we'll be right back to the Firefox issue that inspired this change. :)

wuz: Whoops! Great catch! Wonder why I didn't run into that when I was working on it. 🤷‍♂

assets/css/structures/_grid.scss
```diff
@@ -64,5 +64,10 @@
 
   a {
     color: white;
+
+    &:focus {
+      background-color: pink;
```
tatianamac, line 69: Could we add this a sass variable with its intended purpose (i.e.,`$lt-background-color`; however you want to develop the naming system is fine with me). so that eventually when we activate dark mode we can easily switch these things globally as a theme? Thank you!

tatianamac:
@wuz Hi! Do you have time/bandwidth/energy to take on these edits?

wuz:
@tatianamac yes! I'll take a look at them tonight!

wuz:
@tatianamac after way longer than I said it would take, this is updated!

PR #109, opened 2020-04-19
A11y fixes: add headings and label lists

### LINKED ISSUE #39: HTML: Consider replacing <section>s with <article>s
[Bruce Lawson wrote a very insightful article about the usefulness of `<article>` over `<section>`](https://www.smashingmagazine.com/2020/01/html5-article-section/). Key points:

1. `<section>` communicates no semantic value for assistive technologies (whereas `<article>` does.
2. Apple’s Reader mode ignores content displayed in `<section>` in favour of content displayed in `<article>` (you can see this in action when you use Safari’s Reader mode on selfdefinedapp.com—only the top definition is displayed.)

He also includes a lovely little simile that made me understand the use of `<article>` better:

>think of `<article>` not just as a newspaper article, or a blog post, but as an article of clothing — a discrete entity that can be reused in another context. So your trousers are an article, and you can wear them with a different outfit; your shirt is an article, and can be worn with different trousers;

This would not be a straight swap of `<section>` for `<article>`, as there are some semantic considerations, given that an article should have a heading (in theory the same goes for `<section>`. Perhaps some of the containers would be better suited as `<section>` or even `<div>` with an `aria-label` attribute (there’s a nice example in the article above.)

I don’t like to get too silly or argumentative about semantics, but in terms of inclusivity, I believe this is an easy win.

Having reviewed the CSS, I also believe no CSS is applied to the `<section>` element, so changing the HTML in this instance should not negatively impact any other part of the site.

smhigley:
Hi all 👋

@dengeist pinged me on a list/heading structure issue a while ago, and I thought I'd add a PR for it. This resolves #39 by making the following changes:
- Add `role="list"` to lists for VoiceOver
- Label lists, since navigating by list could be a helpful screen reader technique on the main page (given that all the terms are in nested lists)
- Make first-level list labels (e.g. "A-E") into h3's
- Add two h2's for the two major content sections on the page
- Remove `aria-label`s where invalid/prohibited: This wasn't part of the issue, but I noticed some `aria-label` attributes on term `<span>` and link tags. Since `aria-label` isn't supported on a `<span>` I removed those, and even on the link, it violates [Label in Name](https://www.w3.org/WAI/WCAG21/Understanding/label-in-name.html) in most cases

I did a quick check in common Windows & Mac screen readers, let me know if it looks good and the changes make sense!


ovlb:
@smhigley Thanks so much :) One question: the the label on the spans was something I … messed up. My idea behind this was to give the full name of the sub terms as only the subterm title might be confusing at times. Do you think that might be an issue or was it too defensive from my side?

smhigley:
@ovlb I think you're onto something there, even if I think `aria-label` isn't the best solution. I agree the subterms a bit confusing sometimes (I had that same thought even before seeing the `aria-label`s), but I think that confusion is pretty equitable between visual experience and screen reader-only experience.

I know this belongs in an entirely separate issue, but how do you feel about the idea of having the full name of subterms always displayed? There are a lot of cases where I could read the sub-term only because I already knew it (e.g. "internalized oppression" and not "oppression internalized"), and others where I only discovered the actual term because I tested with a screen reader (e.g. "First Pride"). I imagine in general people might have a range of prior familiarity with terms coming in, and removing some of the puzzle of reading the terms could help more than just screen reader users.

tatianamac:
@smhigley @ovlb All this sounds fine re: naming. If the sub-terming is too confusing in some areas visually and auditorily, we should just remove it in those areas. When we expand interlinking, that will solve the issue the sub-terming was seeking to solve.

tatianamac:
@smhigley I realised that I left the last comment without leaving clear direction for next steps. Are you wanting to make adjustments and push edits to this PR or shall I review it as is?

smhigley:
@tatianamac as-is would be perfect 😃 .

Thank you!

PR #133, opened 2020-05-14
Add "temperature check" alt to "sanity check"

lynncyrin:
## Motivation

"temperature check" is the alternative I've been using personally, and I think it adds some minor value to have it here!

## Usage Comparison

"temperature check" is very similar to the existing "smoke test" alternative, but I think it's different enough to callout separately. For example, here's two distinct ways that I would use the two phrases:

- I stood up a quick PR to fix a bug I found, can you give it a _temperature check_?
- The code just pushed to prod has a very large diff, can we run a _smoke test_ on it?

## Reviewer Guidance

"temperature check" and "smoke test" are close enough in meaning that this might not be a valuable addition. So I wouldn't be offended if this PR wasn't accepted 🙂

### DIFF
--- a/11ty/definitions/sanity-check.md
+++ b/11ty/definitions/sanity-check.md
@@ -14,8 +14,8 @@ alt_words:
   - initial check
   - quick check
   - sense check
-  - temperature check
   - smoke test
+  - temperature check
   - soundness check
   - spot check
 reading:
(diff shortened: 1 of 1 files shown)


tatianamac:
@lynncyrin I agree! I'm wondering about the prospective value of adding descriptions to "alt words" It feels excessive to invite these "subterm suggestions" as full blown definitions (as that's outside of the general scope of this dictionary, but could be beneficial here. Something I migiht open a ticket on. 🤔

Thank you for contributing and bringing your nuanced perspective! Really appreciate it! 🤗

PR #138, opened 2020-05-16
definitions/gender-pronouns: Fix onwards link to 'pronouns'

issyl0:
- First of all, thanks for building SelfDefined! <3
- The pronouns link was missing an 's' on 'definitions/...' and gave a 404.
- I searched for other occurrences of this and didn't find any. :tada:

### DIFF
--- a/11ty/definitions/gender-pronouns.md
+++ b/11ty/definitions/gender-pronouns.md
@@ -4,7 +4,7 @@ slug: gender-pronouns
 defined: true
 ---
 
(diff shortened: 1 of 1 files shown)


tatianamac:
Also! Is it okay with you if I thank you from the Self-Defined twitter account? If yes, what Twitter handle should I tag?

issyl0:
Totally fine! Same handle as here: https://twitter.com/issyl0

PR #147, opened 2020-05-21
Add bierasure definition

TartanLlama:


tatianamac:
Thank you for contributing this, @TartanLlama ! Is it okay if I thank you from the Self-Defined account when I share this?

PR #148, opened 2020-05-21
minoritised definition, minor grammar and punctuation changes

katesowles:
- changed "have grant less power" to "have been granted less power" for readability/grammar
- added period/full-stop to the end of the Benefits section


tatianamac:
Thank you, @katesowles ! May I thank you from the Self-Defined account when I do my next summary roundup?

PR #151, opened 2020-05-22
Feature/rss feed

### LINKED ISSUE #149: Introduce RSS/Atom feed
- allows folks to subscribe and keep up to date with new additions
- to check: curious if RSS/Atom can be setup to push edits/updates as well

And of course 11ty has a plugin for this: https://www.11ty.dev/docs/plugins/rss/

katesawyerdev:
- introduced the [`@11ty/eleventy-plugin-rss` plugin](https://github.com/11ty/eleventy-plugin-rss)
- copied over the [sample feed template](https://www.11ty.dev/docs/plugins/rss/#sample-atom-feed-template)
- for easy access, there is a link to the feed in the "ways to help" CTA box (as well as a <link> tag in the <head>
Screen Shot 2020-05-21 at 4 45 38 PM
- configured the sample feed template for the Self Defined app structure
- removed redundant `---` i noticed in the `-misia` definition because it rendered an `<hr>` that no other posts had, and XML was not happy
- confirmed no linter errors
- confirmed all tests pass

## please note:
- url for site root (https://selfdefined.app) is hardcoded via the metadata, so testing on localhost requires swapping `https://selfdefined.app` for `localhost:[port]` for testing feed links
- XML was super picky and the validator lit up on the rendered post content, despite a `type="html"` attribute on the `<content>` tag. I played around with using [CDATA tags](https://stackoverflow.com/questions/4412395/is-it-possible-to-insert-html-content-in-xml-document), which then surfaced other nit-picky XML validation errors. I chased this down for a while, before just backing out. I'm open to tips and recommendations on the XML if anyone has ideas. despite the validator warnings/errors, I can confirm the feed still works in a feed reader (shown below using Reeder for macOS)
Screen-Recording-2020-05-21-at-4

### DIFF
--- a/.eleventy.js
+++ b/.eleventy.js
@@ -1,6 +1,7 @@
 const definitionPermalink = require('./11ty/helpers/definitionPermalink');
 const renderDefinitionContentNextEntries = require('./11ty/shortcodes/renderDefinitionContentNextEntries');
 const findExistingDefinition = require('./11ty/filters/helpers/findExistingDefinition');
+const pluginRss = require('@11ty/eleventy-plugin-rss');
 
 module.exports = function(config) {
   // Add a filter using the Config API
@@ -37,6 +38,8 @@ module.exports = function(config) {
   config.addFilter('postInspect', function(post) {
     console.log(post);
   });
+
+  config.addPlugin(pluginRss);
 
   config.addShortcode('definitionFlag', (flag) => {
     const cleanText = new Map([
--- a/11ty/_data/metadata.json
+++ b/11ty/_data/metadata.json
@@ -1,6 +1,7 @@
 {
   "title": "Self-Defined",
   "url": "https://www.selfdefined.app/",
+  "feedPermalink": "/feed.xml",
   "description": "A modern dictionary about us. We define our words, but they don't define us.",
   "author": {
     "name": "Tatiana & the Crew",
--- a/11ty/_data/metadata.json
+++ b/11ty/_data/metadata.json
@@ -5,6 +5,7 @@
   "description": "A modern dictionary about us. We define our words, but they don't define us.",
   "author": {
     "name": "Tatiana & the Crew",
+    "name_safe": "Tatiana &amp; the Crew",
     "email": "info@selfdefined.app"
   }
 }
--- a/11ty/_includes/layouts/base.njk
+++ b/11ty/_includes/layouts/base.njk
@@ -13,6 +13,7 @@
     >
     <link rel="stylesheet" href="https://use.typekit.net/qlo3dpu.css" rel="preload"/>
     <link rel="stylesheet" href="{{ '/assets/css/base.css' | url }}">
+    <link rel="alternate" type="application/atom+xml" href="{{ metadata.feedPermalink | absoluteUrl(metadata.url) }}"/>
     {% block pageStyles %}
     {% endblock pageStyles %}
   </head>
--- a/11ty/definitions/-misia.md
+++ b/11ty/definitions/-misia.md
@@ -16,7 +16,7 @@ reading:
   - text: "Anti-Oppression: Anti-Fatmisia"
     href: https://simmons.libguides.com/anti-oppression/anti-fatmisia
 ---
----
+
 from Greek for hate or hatred
 
 ## Use
--- a/11ty/definitions/derpy.md
+++ b/11ty/definitions/derpy.md
@@ -30,7 +30,7 @@ derp, derp face, herp derp, hurr durr
 
 Derpy is very commonly used as an adjective to describe someone's clumsy or foolish behavior, occasionally that of the speaker themselves. Sometimes it is used as a direct replacement of the [r-word](https://www.selfdefined.app/definitions/r-word/).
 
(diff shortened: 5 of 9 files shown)


tatianamac:
This is awesome! Thank you, @katesowles!

Two questions:
1. Did you already test the generated .rss feed through a validator? I know we had some back and forth awhile back about `article`.
2. I wonder if we could put this in the impromptu footer (where "Documentation" currently exists?), e.g., "Documentation · RSS Feed". I worry the "Ways to Contribute" place isn't quite the right place and getting a little cluttered.

ovlb:
@katesowles This is such a wonderful addition to the site. Thank you! 

I did some testing locally and came up with these points:

I ran the resulting feed through the [W3C Feed Validator](https://validator.w3.org/feed/check.cgi), and it’s valid. 

The only warning that should be fixed is regarding the use of `metadata.author.name_safe`

> line 10, column 36: name should not contain HTML: &amp;

The [recommended fix is described in the documentation](https://validator.w3.org/feed/docs/warning/ContainsHTML.html).

(Side note: I believe the `article` discussion we had is independent of the feed use case here, as the latter uses `post.templateContent` which is the rendered markdown, provided by Eleventy)

I also added the feed to Reeder and what struck me as a bit weird is that all definitions have a publishing date which is the last build time, even though `post.date` and the date in the feed are looking okay. I have a similar problem with the feed for my personal site, which I added after seeing this PR (sometimes you just need a bit of inspiration 💞). I will try to make sense of it.

Just came to mind: Our word collections are sorted alphabetically, as we render everything in alphabetical order. The items in the feed should be ordered chronological. I think we might need another collection for this? I took a look at [the source of the `rssLastUpdatedDate` filter](https://github.com/11ty/eleventy-plugin-rss/blob/master/.eleventy.js#L12), this uses the date of the last item in the collection, which leads to confusing results in our (alphabetical) case.

tatianamac:
@ovlb , could it be related to this? https://www.11ty.dev/docs/dates/

ovlb:
@tatianamac Thought so too. But it is rather related to Reeder and Atom Feeds. Apparently Reeder uses the subscription date and time for the initial feed display (interestingly it does not do so for «real» RSS feeds, that’s what tripped me up). 

I deployed my feed (where I had the same problem) to my dev instance. In Feedly everything looks fine. You can check for yourself by using https://11ty.owlish.dev/text/feed.xml

Excuse the confusion. 🙈 

So, I think the updated date issue is the only thing that’s really an issue.

katesowles:
Thank you both for your reviews! I’ll get these changes back to you shortly 😁

katesawyerdev:
Sorry for taking so long to get back to this!

1. I moved the "subscribe" link to the footer with the new text of "RSS Feed" as requested.

2. The validation issue with the `&` should be cleared up by using &amp; in `name_safe`. It shows as passing validation for me using `https://validator.w3.org/feed/check.cgi`:

Screen Shot 2020-06-05 at 7 53 51 PM

Screen Shot 2020-06-05 at 7 54 26 PM

3. About the post date issue... I see what y'all are saying about potentially needing a new collection for sorting these posts by date. But even with that, the post dates all showing the one date of 5/21/2020. From what I can tell wit ha bit of tinkering, I think this may have to do with the fact that 5/21 is when I cloned the repo, which is what I think Eleventy sees as the post/page creation date (which is the default post.date value if one is omitted from the front matter) -- instead of the date the file was added to the repo by the original contributor.

Screen Shot 2020-06-05 at 7 42 02 PM

Presumably the same single-dates showing for you two are the dates you cloned the repos? And then the page dates for files you've created since then? So it seems to me this is different on a per-environment basis (so production should have the ACTUAL dates of creation, right?)

Does that make sense how I explained it? And since y'all have tons more experience here than I do, does that seem right with what all you know about 11ty? I'd be curious to see how dates in production are shown with the rest of the changes I've added here tonight if y'all are up for it?

Otherwise, I'm not sure how to get around this date issue because adding `date: Last Modified` to each post shows today's date since now that I've added the frontmatter, the document has been formatted -- but that's not representative of the date(s) y'all are looking for either... And even adding `date: Created` still shows the same as having no date defined in the frontmatter (makes sense since Created is the default if the value is omitted. So, short of hardcoding a date in each page, I'm not sure how to get around this particular issue.

Let me know if I've misunderstood your messages above! Being 2-weeks removed from this (completely my bad! sorry it has taken this long!), my brain is admittedly a bit fuzzy and not terribly confident in this assessment.

tatianamac:
Thank you, @katesowles ! Do you mind reopening this repo? Migrating selfdefined onto its own org and renaming`master` to `prod` closed all of the issues. 🙃 I'm sorry for the extra work!

katesawyerdev:
@tatianamac Ahh! A rename would explain it! Do you mean reopening the **PR** (doesn't appear I have these permissions)? Or **recloning** the newly-renamed repo and post a new PR to it? Or am I completely lost here? 😆

tatianamac:
Yes, the latter. I did a lot of Things I've Never Done Before™ today. Thank you! 😅

katesawyerdev:
Done! https://github.com/selfdefined/web-app/pull/192

PR #156, opened 2020-05-23
Feature/content warning guidelines

MatthewShields:
Addition of content warning guidelines elements to the definition page template.

Warning pages:

image

image

Avoid pages:

image

image

### DIFF
--- a/11ty/_includes/components/content-warning-flag.njk
+++ b/11ty/_includes/components/content-warning-flag.njk
@@ -1,0 +1,15 @@
+{%- if
+  flag and
+  flag.text and
+  (flag.level == 'avoid') -%}
+  <div class="u-margin-bottom-double content-flag">
+    <p class="small">There is a content warning against this term, we would recommend <a href="#content-warning-guide">reading the guidance</a> on warning other people before using this term in any context.</p>
+  </div>
+{%- elseif
+  flag and
+  flag.text and
+  (flag.level == 'warning') -%}
+  <div class="u-margin-bottom-double content-flag">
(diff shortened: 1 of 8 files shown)


tatianamac:
@MatthewShields ! OMG! AMAZING! 

I think I confused you a bit here, so I'm sorry for that. The content warning should continue and maintain the top styling. In the future, when we build out warnings, I think your secondary styling makes sense (for example, when we discuss slurs).

Is it possible to only apply the top styling for now? Thank you!

Also, could you please remove the [ ] around the word? Thank you!

Once this is merged in, is it okay if I thank you from the Self-Defined Twitter?

MatthewShields:
@tatianamac No problem at all, I'm really enjoying my first contributing experience :)

I am however really sorry that I am going to have to ask just to make sure what it is you are after in terms of the styles just to make sure that I am really clear. (Sorry I will get better at this!)

For the top styling are you wanting:

A:
image

B:
image

For the bottom styling are you wanting:

A:
image

B:
image

C:
Remove styling

I have committed your other suggestions too - love that you can do that directly from the reviews! We don't use github but bitbucket server at work so thats a new one to me 🥳

Also that would be grand if you posted on Twiiter - thanks!

tatianamac:
B on both, please and thank you! No apologies, I appreciate you being thorough.

And yes, that feature is so, so nice! I try to only use it when it's a noodle-y semantic thing and not micromanaging someone's code!

I'm very excited to promote this feature as it ties in with a definition I have pending for suicide (which definitely requires a content warning), so I'll be able to promote both in short order.

MatthewShields:
Great in which case I believe that this PR is ready to be merged :)

I'm really happy to have helped even in this small way.

Please let me know if there is anything further that you need me to do (I think I have everything on my side as I have done the PR to master)

Issue #159, opened 2020-05-24
Broken links in preferred pronoun


oakthielbar:
The links to the definitions for pronoun and neopronouns are both broken on the preferred pronoun definition page. This appears to be due to a missing preceding forward slash.

I'm working on establishing a PR for this now.

PR #160, opened 2020-05-24
Fix broken links in preferred pronoun definition

oakthielbar:
Addresses #159 

Straightforward fix to broken links in the preferred pronoun definition.


tatianamac:
Thank you, @oakthielbar !

PR #161, opened 2020-05-25
created sample content word template.

lkopacz:


tatianamac:
OMG, this is so helpful—thank you, @lkopacz ! Once this is updated to align with the main branch and merged in, would it be okay with you if I thanked you from Self-Defined's Twitter account?

ovlb:
Isn’t the «Read also» headline redundant with the reading part in the front matter?

And should we add this to the docs pages and link to it in the contributing guidelines?

tatianamac:
@ovlb Yes, good point about the read more (as a separate note, for other languages, we have to rely on this so we can translate the heading, but point taken for English and will adjust). I was going to open a ticket tonight about updating the docs with more "How to write a definition" content, which will provide further context for this template.

lkopacz:
> OMG, this is so helpful—thank you, @lkopacz ! Once this is updated to align with the main branch and merged in, would it be okay with you if I thanked you from Self-Defined's Twitter account?

Absolutely!!!

lkopacz:
> Isn’t the «Read also» headline redundant with the reading part in the front matter?
> 
> And should we add this to the docs pages and link to it in the contributing guidelines?

Oh this is a good point!

lkopacz:
@ovlb @tatianamac As I'm remembering why I put the heading there, the reason I put a "Read Also" heading was for definitions that are related. What are your thoughts?

tatianamac:
> @ovlb @tatianamac As I'm remembering why I put the heading there, the reason I put a "Read Also" heading was for definitions that are related. What are your thoughts?

Yes! I think you're right. I'm wondering if within the template if we commented out some guidance terminology; would that be helpful? We can also consider this a future feature and not part of this PR (I can open an issue once we've wrapped for the contribution side of this work  that can include this). 

```
[Read also]
/*  Here link related words that are defined in the dictionary */
```

ovlb:
I had a third think about the read also section while reviewing the issue template, porting my comment for transparency reasons:

> Thinking about for the third time: I think we should move this into front matter and render this via a shortcode or loop, as we have the ability to automatically link words. That ability will be really helpful just in case the link structure changes.
> Plus, we could also link words that will be defined, and the links will be added automatically. As a practical example, there’s this whole currently undefined complex of sexual health terms (PrEP, Undetectable, Viral Load …). By automating the linking we could define the list in all definitions and one by one they get linked without us having the edit n+1 files.

tatianamac:
@lkopacz When you change PR names, they auto-close the PR. Do you mind reopening this PR so we can finish it out? Thank you!

PR #172, opened 2020-05-29
🏡 Updates Packages

ovlb:
This also fixes all but 5 sec vulnerabilities. There is one high left in `http-proxy`, but the exploit does not apply to our project, so I think it’s save to wait for Eleventy to fix the version in their dependencies.


tatianamac:
Thank you, @ovlb ! I think this is very helpful. I agree with the approach to wait until Eleventy updates wrt `http-proxy`

PR #175, opened 2020-05-29
⚒️ Adds "Word definition" template issue

tatianamac:
The hope here is to ease the burden of writing a term and provide a template for folks who aren't able to submit a PR.

### DIFF
--- a/.github/ISSUE_TEMPLATE/word-definition.md
+++ b/.github/ISSUE_TEMPLATE/word-definition.md
@@ -1,0 +1,55 @@
+---
+name: Word definition
+about: Template to define undefined word
+title: "➕ Propose definition for [word]"
+labels: Type · Writing, Type · Definitions
+assignees: ''
+
+---
+
+title: definition
+slug: slug
+speech: 
+alt_words:
+  - alt word
+  - alt word
+  - alt word
+reading:
+  - text: 'Link Title'
+    href: https://google.com
+  - text: 'Link Title'
+    href: https://google.com
+---
+
+Definition goes here. Below are some headings you can use in the definition. Generally 2-3 sentences max; include colloquial definitions as relevant.
+
+## Issues
+/* When the term is problematic, describe what issues there are. If term is not problematic, this can be omitted. */
+
+## Impact
+/* When the term is problematic, describe what impact there is. If term is not problematic, this can be omitted. */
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/word-definition.md
```diff
@@ -0,0 +1,55 @@
...
+## Appropriate Usage
+/* When term has appropriate and inappropriate usage; common for medical or culturally appropriative terms */
+
+## Read also
+/* When a term has related words that are defined */
+
+- [definition](/definitions/[word])
+- [definition](/definitions/[word])
+- [definition](/definitions/[word])
+
```
ovlb, line 53: Thinking about for the third time: I think we should move this into front matter and render this via a shortcode or loop, as we have the ability to automatically link words. That ability will be really helpful just in case the link structure changes. 

Plus, we could also link words that _will_ be defined, and the links will be added automatically. As a practical example, there’s this whole currently undefined complex of sexual health terms (PrEP, Undetectable, Viral Load …). By automating the linking we could define the list in all definitions and one by one they get linked without us having the edit n+1 files.

This also applies to #161

.github/ISSUE_TEMPLATE/word-definition.md
```diff
@@ -0,0 +1,55 @@
...
+title: definition
+slug: slug
+speech: 
+alt_words:
+  - alt word
+  - alt word
+  - alt word
+reading:
+  - text: 'Link Title'
+    href: https://google.com
```
ovlb, line 19: Do we want to use placeholder links that are not FAANG here?

ovlb (commented):
@tatianamac Had a look and this is a superb step. I had some hopefully helpful ideas ranging from very small to a tad more impactful.

.github/ISSUE_TEMPLATE/word-definition.md
tatianamac: I think we can make them generic like https://www.link.com

tatianamac:
@lkopacz  Because I ported the repo over to a new org and renamed the main branch from master (which was problematic) to prod, I think you'll need to reopen it and make a PR to the new main branch (prod). Reach out to the Slack channel for help. 🙏🏽

PR #182, opened 2020-06-01
Issue/show definition in social

### LINKED ISSUE #179: ⚒️ Show definition in unfurled social views
## Current Behaviour
Right now the description says "Definition of [word] in Self-Defined, a modern dictionary about us:

image

## Desired Behaviour
Instead, it'd be great if the social card/unfurled view l could include the definition

kathryngraysonnanz:
This is a proposal to fix the way links preview in social (issue link: https://github.com/tatianamac/selfdefined/issues/179) 

I'd like to suggest adding a new front matter section to the definitions – "socialPreview". This will allow the user to draft a short version of the definition that will appear in the social previews. I experimented with adapting the existing content by taking sections of the definition, but found that:
- the actual definition is often far too long and 
- it doesn't always lead with the information that might be most relevant in a social preview (such as warnings about whether or not to use the word in conversation). 

Rather than try and make assumptions about how the author may want their definition to appear in preview, or try and auto-parse through the content, I suggest giving the author the ability to write the preview copy exactly how they like it. 

I've also added the flag level (warning, avoid, etc.) _before_ the preview, since it seems like there are some definitions that really require that important context. 

Here are a couple screenshots that show how the previews appear with these edits: 
Screen Shot 2020-06-01 at 4 03 04 PM
Screen Shot 2020-06-01 at 4 03 57 PM


11ty/_includes/layouts/base.njk
```diff
@@ -7,10 +7,29 @@
     {# Use title with path, or append a space to the page title to avoid collpasing with the meta title #}
     {% set pageTitle = titleWithPath or title + ' ' or '' %}
     <title>{{ pageTitle + metadata.title }}</title>
+
+    <meta
+      name="twitter:description"
```
kathryngraysonnanz, line 12: twitter, of course, doesn't use the open graph standard because they're ~ special ~

11ty/definitions/-misia.md
```diff
@@ -3,6 +3,7 @@ title: -misia
 slug: -misia
 speech: noun
 defined: true
+socialPreview: from Greek for hate or hatred
```
kathryngraysonnanz, line 6: Added the `socialPreview` front matter to just a few existing definitions, for testing and review purposes. If we adopt this approach, I'll go back and add it to the others.

MatthewShields:
I really like this feature! 🙂 I think this makes complete sense for when someone may be using a definition in something like a slack channel to point out to someone why their language is inappropriate.

Just to check though, on the definitions where the social preview text hasn't been provided does it fall back to using the meta description or nothing at all? (If twitter:title="").

@tatianamac do you think that there would be a need to provide guidance about writing a social preview and the language included in it? as it may be displayed directly to the user (rather then having been seen after choosing to go through to the page)

ovlb:
Hey Kathryn,

Great that you were able to get this up!

I especially like the addition of the warning level to the beginning, to provide context.

Some things that came to mind:

- We should give the front matter part a generic name like `excerpt` (we could also use this in a «Definition of the day» widget or something, that isn’t _directly_ social). 
- If the excerpt isn’t set we need to ensure there’s a generic fallback (like [the current one](https://github.com/tatianamac/selfdefined/blob/master/11ty/_includes/layouts/definition.njk#L4))
- `og:description`, `twitter:description` and `description` all need to have the same value

More on the editorial site: As `description` is also used by search engines to show the preview and e.g. in a sharing context in Facebook or Twitter the length of the string is more limited than in Slack we would need to ensure to a max length, or that the important information is at the beginning.

kathryngraysonnanz:
@MatthewShields @ovlb 

Really helpful notes – thank you! I've updated the PR to now: 
- include fallbacks for the fields that may not be filled out  
- rename the 'socialPreview' field to 'excerpt' 
- make sure all the descriptions are consistent 
- add a character limit and "see more" note to the end of the Twitter preview, which supports fewer characters than the others (based on @tbrasington 's linked PR) 

I do also agree that we should provide guidance on writing these excerpts, but that feels like a documentation issue and not a code issue to me, IMHO.

MatthewShields:
Great stuff! 🙂 

I would definitely agree that guidance would be a documentation thing rather then the code itself.

Good luck with the checks, I'm sensing that we are in different timezones but if there anything I can help with please just let me know.

kathryngraysonnanz:
Alright, updated this to fix the failing checks. I think this should be good to go now :) LMK if you guys see anything else that needs adjustment. And thanks for all the helpful feedback!

tatianamac:
@kathryngraysonnanz  Because I ported the repo over to a new org and renamed the main branch from master (which was problematic) to prod, I think you'll need to reopen it and make a PR to the new main branch (prod). Reach out to the Slack channel for help. 🙏🏽

PR #191, opened 2020-06-04
Adding preconnect and dns-prefetch hints in header for TypeKit hosts.

malchata:
This PR adds `preconnect` and `dns-prefetch` headers to the document `<head>` for TypeKit hosts. These hints will establish a connection to the both use.typekit.com and p.typekit.com as soon as the browser is able to instead of at discovery time. This will not yield much of an improvement for loading content from use.typekit.com, but a connection to p.typekit.com at current is not initiated until it is discovered in the CSS resource served from use.typekit.com:

WebPageTest waterfall for selfdefined.app without resource hints

By adding these resource hints, we should be able to move DNS resolution, connection negotiation, and TLS negotiation further up in the process, ensuring less delay in both CSS and font loading from TypeKit. I can run a followup test once the changes are pushed.

### DIFF
--- a/11ty/_includes/layouts/base.njk
+++ b/11ty/_includes/layouts/base.njk
@@ -1,9 +1,12 @@
 <!doctype html>
 <html lang="en">
   <head>
+    <link rel="preconnect" href="https://use.typekit.net/" crossorigin />
+    <link rel="dns-prefetch" href="https://use.typekit.net/" />
+    <link rel="preconnect" href="https://p.typekit.net/" crossorigin />
+    <link rel="dns-prefetch" href="https://p.typekit.net/" />
     <meta charset="utf-8">
     <meta name="viewport" content="width=device-width, initial-scale=1.0">
-
     {# Use title with path, or append a space to the page title to avoid collpasing with the meta title #}
     {% set pageTitle = titleWithPath or title + ' ' or '' %}
     <title>{{ pageTitle + metadata.title }}</title>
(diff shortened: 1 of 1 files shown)


11ty/_includes/layouts/base.njk
```diff
@@ -1,17 +1,20 @@
...
     <meta name="viewport" content="width=device-width, initial-scale=1.0">
-
     {# Use title with path, or append a space to the page title to avoid collpasing with the meta title #}
     {% set pageTitle = titleWithPath or title + ' ' or '' %}
     <title>{{ pageTitle + metadata.title }}</title>
     <meta
       name="description"
       content="{{ renderData.description or description or metadata.description }}"
     >
-    <link rel="stylesheet" href="https://use.typekit.net/qlo3dpu.css" rel="preload"/>
```
malchata, line 14: The `preload` hint here is removed, as this is not the preferred format for `preload`. In this case, that would be an entirely separate `<link>` tag like this:

```html
<link rel="preload" as="style" href="https://use.typekit.net/qlo3dpu.css" crossorigin />
```

Even so, this hint would be so close to the stylesheet in the <head> that it would have no meaningful effect. We can restore if you would prefer.

11ty/_includes/layouts/base.njk
```diff
@@ -1,17 +1,20 @@
 <!doctype html>
 <html lang="en">
   <head>
+    <link rel="preconnect" href="https://use.typekit.net/" crossorigin />
+    <link rel="dns-prefetch" href="https://use.typekit.net/" />
```
malchata, line 5: This may seem redundant, but there's a reason for each of these hints:

1. `preconnect` will establish a complete connection to the origin specified in `href`.
2. Not all browsers support `preconnect`, but a larger number of browsers support `dns-prefetch`, which performs the DNS lookup rather than establish a connection to the remote server. Not as effective, but still gives something of a boost to older browsers.

tatianamac:
@malchata  Because I ported the repo over to a new org and renamed the main branch from master (which was problematic) to prod, I think you'll need to reopen it and make a PR to the new main branch (prod). Reach out to the Slack channel for help. 🙏🏽

malchata:
Gotcha. Thanks for the heads up!

PR #192, opened 2020-06-06
Feature/rss feed

### LINKED ISSUE #149: Introduce RSS/Atom feed

katesawyerdev:
# Related to PR #151 

Sorry for taking so long to get back to this!

1. I moved the "subscribe" link to the footer with the new text of "RSS Feed" as requested.

2. The validation issue with the `&` should be cleared up by using &amp; in `name_safe`. It shows as passing validation for me using `https://validator.w3.org/feed/check.cgi`:

Screen Shot 2020-06-05 at 7 53 51 PM

Screen Shot 2020-06-05 at 7 54 26 PM

3. About the post date issue... I see what y'all are saying about potentially needing a new collection for sorting these posts by date. But even with that, the post dates all showing the one date of 5/21/2020. From what I can tell wit ha bit of tinkering, I think this may have to do with the fact that 5/21 is when I cloned the repo, which is what I think Eleventy sees as the post/page creation date (which is the default post.date value if one is omitted from the front matter) -- instead of the date the file was added to the repo by the original contributor.

Screen Shot 2020-06-05 at 7 42 02 PM

Presumably the same single-dates showing for you two are the dates you cloned the repos? And then the page dates for files you've created since then? So it seems to me this is different on a per-environment basis (so production should have the ACTUAL dates of creation, right?)

Does that make sense how I explained it? And since y'all have tons more experience here than I do, does that seem right with what all you know about 11ty? I'd be curious to see how dates in production are shown with the rest of the changes I've added here tonight if y'all are up for it?

Otherwise, I'm not sure how to get around this date issue because adding `date: Last Modified` to each post shows today's date since now that I've added the frontmatter, the document has been formatted -- but that's not representative of the date(s) y'all are looking for either... And even adding `date: Created` still shows the same as having no date defined in the frontmatter (makes sense since Created is the default if the value is omitted. So, short of hardcoding a date in each page, I'm not sure how to get around this particular issue.

Let me know if I've misunderstood your messages above! Being 2-weeks removed from this (completely my bad! sorry it has taken this long!), my brain is admittedly a bit fuzzy and not terribly confident in this assessment.


tatianamac:
Thank you, @katesowles . @ovlb, could I rely on you to work on this with Kate? We can also call in the Slack group to see if anyone has good ideas.

(Kate, I don't think I had extended an invite to you to our [Slack group](https://join.slack.com/t/selfdefined/shared_invite/zt-en1u6xn8-CumcehUfprC0GH5lamb1bg). No pressure since I know everyone is in too many slacks, but if you'd like help and/or to contribute again in the future, we'd love to have you!

katesawyerdev:
@ovlb @tatianamac thank you both! I'd be honored to add this slack to my sidebar!

tatianamac:
@ovlb will you let @katesowles know if you have any edits?

ovlb:
@tatianamac Reviewing it now, have been off the grid for a few days.

ovlb:
>  I'd be curious to see how dates in production are shown with the rest of the changes I've added here tonight if y'all are up for it?

Yah. I would say we should test in some popular clients once we merged this to prod and see if issues come up. :) As mentioned in the old PR (https://github.com/selfdefined/web-app/pull/151#issuecomment-633247377) some things from which I thought they were issues turned out to be none. :)

.eleventy.js
```diff
@@ -157,6 +160,17 @@ module.exports = function(config) {
...
 	})
 
+  config.addCollection('definedWordsChronological', (collection) => {
+    return collection
+      .getFilteredByGlob('./11ty/definitions/*.md')
+      .filter((word) => word.data.defined)
+      .sort((a, b) => {
+        if (a.date > b.date) return 1
+        if (a.date < b.date) return -1
+        return 0;
```
ovlb: The sort should be the other way around so that we have the newest item first.

Witht this the updated timestamp looks like `<updated>2020-05-17T15:04:44.465+02:00</updated>`. But the timestamp of the newest item is `<updated>2020-06-10T18:02:15.575+02:00</updated>`.

ovlb (changes requested):
A small change needed regarding the sorting of the array. Once this is done we can merge this. Thanks a ton, Kate.

.eleventy.js
katesawyerdev: Makes sense! Pushing up the change now!

ovlb (approved):
We got a feed. 💞

ovlb:
@katesowles Thanks so much again. As Tatiana always asks, mind if we thank you from our Twitter account as we announce the feature?

katesawyerdev:
@ovlb sure 🙂  thank you (both) for asking!

PR #199, opened 2020-06-09
Add optional category support for alt words.

### LINKED ISSUE #170: ⚒️ Add 1:1 term comparison
## Current Landscape
Currently, alt words are a flat front matter datum. Some words that we're suggesting to avoid can be used in multiple contexts.

For example, dude can be used as an interjection and as a noun. It becomes unclear which alt words correspond with what. See #169 for details.

## Feature Description
Many ways exist to solve this problem, a couple outlined below. Open to other ideas!

### 1️⃣ Allow for basic level descriptions
For example, the front matter could look like this:
```
- alt_word:
  - interjection
     - dang
     - wow
  - person
     - you or y'all
     - friend
```
Here the "levels" are manually entered by the writer and the user has to make the mental connection using context (so in the definition, it might use these same words).

### 2️⃣ Allow for advanced level descriptions
The usage context would be the same as above, but instead of only manually connecting them, figuring out a way to make the connection dynamic. One method would be to connect a key word. So, in the example above, perhaps the word [interjection] could be automatically linked to its corresponding alt-word, such that clicking on it would link you to the alt word anchored section for that word.

sarenji:
## What

In Markdown word definitions, the frontmatter contains an `alt_words` key, which only supports arrays. This PR is a first-pass at expanding this to support categories (as a hash/object). #170 outlines an "advanced level description", but I wasn't totally sure I understood the outline, so I opted for the simpler solution for me.

## Usage

I changed the entry for `dude` as a start. We can incrementally edit the other entries as needed.

Before, we only supported this format:

```yaml
alt_words:
  - wow
  - oh my word
  - yikes
  - y'all
  - friend
  - mate
  - fam
  - folks
  - pal*
  - buddy*
```

Now, we also support categories:

```yaml
alt_words:
  expression of astonishment:
    - wow
    - oh my word
    - yikes
  noun:
    - y'all
    - friend
    - mate
    - fam
    - folks
    - pal*
    - buddy*
```

**I'm happy to change the category names to fit #170.** I pulled these categories from the content, and I also wanted to test that spaces in category names translated to the right HTML `id` :grin:

## Technical details

When rendering `alt_words`, we branch on whether the value is an object (`{}`) or an array (`[]`). If it's an array, then like before, we render the alternatives as a semi-colon-separated list. If it's an object, we take the keys (category name) and the values (either an array of alternatives or another category object), then render the category names while redoing these steps again on the value.

## What does it look like?

Alternatives for the word dude with the implemented changes

**Image description:** The alternative words for "dude" are separated into two categories: "expression of astonishment" and "noun". The categories are bolded, and underneath each category are its associated alternative words in a list separated by semi-colons.

## Caveats

1. I couldn't tell if there was a process for claiming issues or starting discussions about which solution to pursue. In any case, I am not attached to any of this code/design, and I'm happy to change anything, or throw all this away.
1. This only really supports one level of categories; I didn't style for multiple category levels. I think multi-level probably deserves a bigger conversation + more thoughtful styling.
1. I'm not a full-time screen reader user so I did some guesses on the appropriate labeling strategy for the lists and tested it out in practice (with VO). But, I can't say for sure it's the right solution.
1. I did not update documentation. I wanted to get an idea of what y'all are looking for first.

### DIFF
--- a/11ty/_includes/components/definition-content.njk
+++ b/11ty/_includes/components/definition-content.njk
@@ -1,16 +1,40 @@
+{% macro renderAltWords(words, labels) %}
+  {%- if words | isArray -%}
+    {# Array #}
+    <ul
+      class="list-semicolon"
+      {% if labels %}aria-labelledby="{{ labels | join(' ') }}"{% endif %}
+    >
+      {%- for word in words -%}
+        <li>
+          {{ word | linkIfExistsInCollection(collections.definedWords) | safe }}
+        </li>
+      {%- endfor -%}
+    </ul>
+  {%- else -%}
+    {# Object #}
+    {# With class="", the list loses padding via the CSS reset style. #}
+    {# For multi-level category support, we may want an actual class. #}
+    <ul class="">
+      {%- for category, list in words -%}
+        <li>
+          {% set categoryId = ("alt-words-" + category) | replace(' ', '-') %}
+          <strong id="{{ categoryId }}">{{ category }}</strong>
+          {{ renderAltWords(list, labels.concat(categoryId)) }}
+        </li>
+      {%- endfor -%}
+    </ul>
+  {%- endif -%}
+{% endmacro %}
+
 <section class="definition-content">
   <section class="definition-content__content">
     {{ content | safe }}
-    {# <p>{{ alt_words }}</p> #}
   </section>
   {%- if alt_words -%}
     <section class="definition-content__content">
       <h2 id="alt-words">Alt Words</h2>
-      <ul class="list-semicolon" aria-labelledby="alt-words">
-        {% for word in alt_words %}
-          <li>{{ word | linkIfExistsInCollection(collections.definedWords) | safe }}</li>
-        {% endfor %}
-      </ul>
+      {{ renderAltWords(alt_words, ["alt-words"]) }}
     </section>
   {% endif %}
   {%- if reading -%}
--- a/.eleventy.js
+++ b/.eleventy.js
@@ -38,6 +38,10 @@ module.exports = function(config) {
     console.log(post);
   });
 
+  config.addFilter('isArray', function(thing) {
+    return Array.isArray(thing);
+  });
+
   config.addShortcode('definitionFlag', (flag) => {
     const cleanText = new Map([
       [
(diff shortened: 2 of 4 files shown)


tatianamac:
Hi, @sarenji ! Welcome! Thanks for contributing. The PR looks good to me. It addresses the 1️⃣ part of #170 imho. Two requires further discussion as it'll rely possibly on the API or other yet-to-be-built features.

I'm tapping a few folks to review the code.

If you'd like, we have a [Slack community](https://join.slack.com/t/selfdefined/shared_invite/zt-en1u6xn8-CumcehUfprC0GH5lamb1bg) that is pretty tame for folks to get help and discuss terms.

11ty/_includes/components/definition-content.njk
```diff
@@ -1,16 +1,40 @@
...
+        <li>
+          {{ word | linkIfExistsInCollection(collections.definedWords) | safe }}
+        </li>
+      {%- endfor -%}
+    </ul>
+  {%- else -%}
+    {# Object #}
+    {# With class="", the list loses padding via the CSS reset style. #}
+    {# For multi-level category support, we may want an actual class. #}
+    <ul class="">
```
arielsilvestri, line 18: I don't know if there are any plans to go beyond this extra level of nesting, so this may be fine?

arielsilvestri (approved):
I recall that this feature request came out of my attempts to do this, and this is sort of how I envisioned the usage would eventually look. I think this is great!

11ty/_includes/components/definition-content.njk
tatianamac: I tend to agree! I think for right now this is enough. More nesting would make things too complicated for the user imho.

tatianamac:
@sarenji Do you mind resolving the conflicts and then we can merge this in? Thank you!

sarenji:
@tatianamac done!

NB: I had to fix some Markdown linting issues on the digital Blackface article in e3f8af1 before it would let me push my changes. The commit also fixes a couple list issues in production

sarenji:
Sorry about the noise. Other PRs merged to `prod` keep causing conflicts with this branch 😅 I'll keep fixing the conflicts as they come.

tatianamac (approved):
🚀 LGTM

tatianamac:
Thank you, @sarenji ! I think this is good. The merge conflicts weren't your fault. We were having test issues that were causing linting issues that caused a downstream mess, so thank you for your patience! I think the long-term solution is good now, so hopefully you don't encounter more issues.

PR #201, opened 2020-06-10
adds optical allyship

AutoSponge:
It's mostly the same impact section as performative allyship with a direct example of #blackouttuesday.


11ty/definitions/optical-allyship.md
```diff
@@ -0,0 +1,28 @@
...
+    href: https://www.instagram.com/p/BiPDZkbFJFY/?hl=en
+  - text: 'If You Want To Be Anti-Racist, This Non-Optical Allyship Guide Is Required Reading'
+    href: https://www.vogue.co.uk/arts-and-lifestyle/article/non-optical-ally-guide
+  - text: 'What Is Optical Allyship? 3 Ways To Be Actively Anti-Racist'
+    href: https://www.bustle.com/p/what-is-optical-allyship-3-ways-to-be-actively-anti-racist-22956518
+  - text: 'What is optical allyship? Ways you can support the BLM movement apart from spreading hashtags'
+    href: https://www.msn.com/en-ca/news/politics/what-is-optical-allyship-ways-you-can-support-the-blm-movement-apart-from-spreading-hashtags/ar-BB15cJgz
+---
+
+> allyship that only serves at the surface level to platform the 'ally,' it makes a statement but doesn’t go beneath the surface and is not aimed at breaking away from the systems of power that oppress.
```
tatianamac, line 24: Can we rewrite this lead-in so it avoids a circular definition? Perhaps something like "when X happens,"

11ty/definitions/optical-allyship.md
```diff
@@ -0,0 +1,28 @@
...
+
+## Related Words
+
+performative allyship, non-optical allyship
+
```
tatianamac, line 29: It may benefit from adding an `## Examples` section to show a few examples of what this looks like?

Also, what about adding a comparison or relational section about the difference between performative allyship and optical allyship.

11ty/definitions/optical-allyship.md
```diff
@@ -0,0 +1,28 @@
...
+## Impact
+
+Optical allyship detracts from the stories of the oppressed and the impact that oppression has on them in favour of the stories of the oppressors and their intent to help. The cost of optical allyship can also be that the stories of the oppressed people are appropriated, watered down, and inaccurately retold by people who should not be telling those stories. For example, in #blackouttuesday, many rushed to publicly respond without understanding the intent which "derailed the conversation."
```
tatianamac:
```suggestion
Optical allyship detracts from the stories of the oppressed and the impact that oppression has on them in favour of the stories of the oppressors and their intent to help. Optical allyship centres the perspectives of the oppressor. The cost of optical allyship can also be that the stories of the oppressed people are appropriated, watered down, and inaccurately retold by people who should not be telling those stories. For example, in #BlackOutTuesday, many rushed to publicly respond without understanding the intent which "derailed the conversation."
```

tatianamac (approved):
A few non-critical suggestions.

11ty/definitions/optical-allyship.md
AutoSponge: I quoted the definition from the source/coiner. I don't feel comfortable redefining it.

11ty/definitions/optical-allyship.md
AutoSponge: Both good ideas. I'll need to study this a little deeper.

11ty/definitions/optical-allyship.md
tatianamac: @autosponge I don't think we're seeking to define it, but instead adjust the introduction so that it fits the format of a dictionary entry. If you don't feel like you're able to do that, I'm happy to do it.

tatianamac: an alternate approach is to leave the definition (perhaps prepending with "a form of allyship" and then to define allyship as a separate term. My main piece of feedback here is that it needs to fit into the dictionary definition framework. I hear your points about wanting to ensure the dictionary definition is accurate to the coiner and agree there.

AutoSponge: I added 'a form of allyship.' as the first line.

tatianamac:
```suggestion
> a form of allyship that only serves at the surface level to platform the 'ally,' it makes a statement but doesn’t go beneath the surface and is not aimed at breaking away from the systems of power that oppress.
```

tatianamac:
@AutoSponge Is this ready for review?

tatianamac:
@AutoSponge  I think there's just one suggestion left above then we can update the branch and merge in.

tatianamac (approved):
🚀 LGTM

tatianamac:
@AutoSponge When we post it on Self-Defined's Twitter, would it be okay with you if we attributed you?

PR #209, opened 2020-06-12
💅 Fix lint error in digial blackface

ovlb:


ovlb:
@tatianamac Do you now why the CI test is only showing up for some PRs? It ran in #204 but isn’t showing up here. 😕  That’s probably also the reason why the lint error got merged in the first place.

tatianamac (approved):
🚀 LGTM

tatianamac:
> @tatianamac Do you now why the CI test is only showing up for some PRs? It ran in #204 but isn’t showing up here. 😕 That’s probably also the reason why the lint error got merged in the first place.

To close the loop here, it's because we only had the test run for push events, not for pull events. Thanks to a PR #212 by @bnb that is now fixed. Let me know if you still encounter issues. ✨

PR #216, opened 2020-06-13
Fluid type scale

### LINKED ISSUE #47: 🔡 Develop more fluid type system
Current type system does not necessarily scale well. Intentionally developing a system that does upon the current typography would be great!

codeAbul:
- Creates a fluid type scale.
- Provides a convenient config map and a mix-in to programatically generate all the css custom properties required for creating and using a type scale.  The methods employed are inspired the following sources( https://every-layout.dev/rudiments/modular-scale/ ). Thanks to @tbrasington and @ovlb for the help and inspiration! 
- Replaced all 'raw' font-sizes with fluid scales. Please let me know if the scales need more refining

Further improvements possible:

- [ ] add clamps for every fontsize used ( need to decide with the help of a designer )
- [x] Create scales for line-height too


tatianamac:
Thank you so much for all your work on this, @codeAbul ! 🎉

tatianamac:
Closes #47.

codeAbul:
Thank you @tbrasington for the updates to line-height and refining scales!

tatianamac:
@ovlb Feel free to approve this if you and @tbrasington are good with it!

ovlb:
@codeAbul FYI: I already had a look and overall it looks fine. I’ll write a review that’s worthy of your effort tomorrow, promised!

tbrasington:
The foundation is solid and in a good place to be extended it as the new features get built out. 

As a side note, what may be needed next is guides on pairings of steps and styles.

assets/css/base/_base.scss
```diff
@@ -1,7 +1,10 @@
+@import 'typography';
```
ovlb, line 1: We import this in `css/base.scss`, I think you were adding this here because otherwise the SASS variables you are defining in the typography partial would not be accessible? 

Two ways to mitigate this, both cleaning up the structure a bit:
- Define the SASS variables in `abstracts/variables` (or in `base/_fonts` where the font weights and families are defined)
- Change the SASS variables to custom properties and define them in `base/custom-properties`

assets/css/base/_base.scss
```diff
@@ -1,7 +1,10 @@
...
+
 body {
   border-top: 1rem solid $primary-color;
   font-family: $sans-serif;
-  font-size: pxToRem(20);
+  font-size: var(--step0);
+  line-height: $line-height-heading;
```
ovlb, line 7: Why are you adding `$line-height-heading` to the body element? I see that you are overwriting this in multiple places down below. I would propose adding `$line-height-body` here (because it’s the body) and delete all places were you are setting `$line-height-body` on other elements. Headlines etc. can still have the deviating line-height. This way we let the cascade of CSS work for us.

```suggestion
  line-height: $line-height-body;
```

assets/css/base/_type-scale.scss
```diff
@@ -0,0 +1,74 @@
...
+/// This is the single source of truth
+$fluid-config: (
+  // Typescale ratio for calculating steps
+    typescale-ratio: 1.25,
+  // base font size to use for --step0
+    base-fontsize: 1rem,
+  // adjustment added to base-fontsize in --step0 for scaling
+    viewport-scale: 0.5vw,
+  // number of positive steps (--step1, --step2 etc)
+    // and number of negative steps (--step-1, --step-2 etc)
```
ovlb, line 13: These names are not really clear. Hyphenating is common to separate words, especially in CSS where we don’t use camel case. Can you change these two be either `--step-pos-1` and `--step-neg-1` or for positive use `--step-1` (as this is what we mostly use) and only namespace the negative variables, e.g. `--step-minus-1`

assets/css/base/_typography.scss
```diff
@@ -1,63 +1,57 @@
...
 }
 
 .sub-headline {
-	font-family: $ext-sans;
-	font-size: 1.25rem;
-	font-weight: $bold;
-	letter-spacing: 0.1;
-	margin: .75rem 0;
+  font-family: $ext-sans;
+  font-size: var(--step0);
```
ovlb, line 47: I believe `step-0` is not the correct value here, as `.sub-headline` should be larger then the body text.

assets/css/structures/_definition-content.scss
```diff
@@ -31,17 +31,18 @@
 
     & > p {
       font-family: $sans-serif;
-      font-size: 1.5rem;
+      font-size: var(--step0);
```
ovlb, line 34: I think you changed this here because `1.5rem` made the first paragraph in definition mostly unreadable on mobile?

ovlb (changes requested):
Hey,

Overall this looks good and especially on mobile increases readability. There are some steps we can take to clean up the code before we merge this in. :)

codeAbul:
@ovlb Thank you for the reviews. I shall address them this weekend. Sorry, been busy with stuff! Thanks again

ovlb:
@codeAbul Your pace is the right pace :) Looking forward to your updates!

tatianamac:
@codeAbul Checking in to see if you're still interested and able to close out this PR. As far as I can tell we're very close!

codeAbul:
Hi @tatianamac i am so sorry I have left this out in the wind , thanks to the mess of pandemic and job change. I will work on this tonight. Again sorry for the delay. Hope you and yours are keeping well, as well as one can be in these times.

assets/css/base/_type-scale.scss
codeAbul: I agree. Will update

tatianamac:
@codeAbul : Hello! A no-pressure check to see how you're feeling about this PR (I'm going through all issues). Are you still interested in finishing it out? (It might very well be finished; it's hard for me to tell for certain).

PR #226, opened 2020-06-17
Content templates

lkopacz:


word-template.md
```diff
@@ -0,0 +1,42 @@
+---
+title: definition
+slug: slug
+flag:
+  text: 'Flag Text'
+  level: avoid
+defined: true
```
kathryngraysonnanz, line 7: Hey there! In accordance with this PR (https://github.com/selfdefined/web-app/pull/227) I'd like to suggest the addition of the new "excerpt" front matter  

```suggestion
defined: true
excerpt: A shortened definition to be used in social media previews goes here. 
```

lkopacz: Oh good call! Thanks for calling this to my attention. I missed this as I'm returning to this from a month ago 🙈

tatianamac:
@lkopacz Checking in to see if you're still interested in finishing out this PR! We're here to support in whatever ways we can.

PR #243, opened 2020-06-22
Added preconnect/dns-prefetch hints for Typekit CSS resources.

malchata:
_This PR is a repeat of #191, which was closed (along with other PRs) when the repo was moved. Some of the text here is adapted from that original PR._

This PR adds `preconnect` and `dns-prefetch` headers to the document `<head>` for TypeKit hosts. These hints will establish a connection to the both use.typekit.com and p.typekit.com as soon as the browser is able to instead of at discovery time.

These changes won't yield much of a performance boost for the CSS loaded from use.typekit.com, but currently, the connection to p.typekit.com isn't initiated until it is discovered in the CSS resource served from use.typekit.com, which you can see in this WebPageTest waterfall:

WebPageTest waterfall for selfdefined.app, showing connection latency for Typekit hosts.

By adding these hints, we'll be able to mask the connection latency for TypeKit hosts by performing a connection as early as possible. This should translate into a modest performance improvement which should speed up font loading.


11ty/_includes/layouts/base.njk
```diff
@@ -3,17 +3,20 @@
...
+    <link rel="preconnect" href="https://p.typekit.net/" crossorigin>
+    <link rel="dns-prefetch" href="https://p.typekit.net/">
     {# Use title with path, or append a space to the page title to avoid collpasing with the meta title #}
     {% set pageTitle = titleWithPath or title + ' ' or '' %}
     <title>{{ pageTitle + metadata.title }}</title>
     <meta
       name="description"
       content="{{ renderData.description or description or metadata.description }}"
     >
-    <link rel="stylesheet" href="https://use.typekit.net/qlo3dpu.css" rel="preload"/>
```
malchata, line 14: This is currently not the correct syntax for preloading a stylesheet, so I removed the `rel` attribute here. If we were to preload this stylesheet, it would be in a separate `<link>` tag that looks like this:

```html
<link rel="preload" as="style" href="https://use.typekit.net/qlo3dpu.css" crossorigin>
```

Even so, the CSS is so far up in the head that preloading it will have no practical effect. The main bottleneck in font loading is waiting for the browser to connect to p.typekit.net after it discovers the `@import` in the CSS loaded from use.typekit.net.

11ty/_includes/layouts/base.njk
```diff
@@ -3,17 +3,20 @@
   <head>
     <meta charset="utf-8">
     <meta name="viewport" content="width=device-width, initial-scale=1.0">
-
+    <link rel="preconnect" href="https://use.typekit.net/" crossorigin>
+    <link rel="dns-prefetch" href="https://use.typekit.net/">
```
malchata, line 7: `dns-prefetch` alternatives are provided here for browsers that don't understand `preconnect`, but do understand `dns-prefetch`. It's not as big of a boost as `preconnect`, but it's better than nothing.

11ty/_includes/layouts/base.njk
```diff
@@ -3,17 +3,20 @@
...
     {% set pageTitle = titleWithPath or title + ' ' or '' %}
     <title>{{ pageTitle + metadata.title }}</title>
     <meta
       name="description"
       content="{{ renderData.description or description or metadata.description }}"
     >
-    <link rel="stylesheet" href="https://use.typekit.net/qlo3dpu.css" rel="preload"/>
+    <link rel="stylesheet" href="https://use.typekit.net/qlo3dpu.css">
     <link rel="stylesheet" href="{{ '/assets/css/base.css' | url }}">
-    <link rel="alternate" type="application/atom+xml" href="{{ metadata.feedPermalink | absoluteUrl(metadata.url) }}"/>
```
malchata, line 16: Noticed that some tags had XML-style self closing slashes and some didn't.

11ty/_includes/layouts/base.njk
ericwbailey: I agree. The CSS payload size seems to be small enough to hold off on `media="print"` trickery for now.

ericwbailey (approved):
Looks good to me!

ovlb (approved):
@malchata Thanks for this contribution. Your first, yay! Hope to see you around in the future 🚀

tatianamac:
@malchata Would it be okay if we thanked you when we post about the feature on our Twitter account?

malchata:
Totally up to you, @tatianamac. I'm not in it for the credit, just wanted to lend a hand.

PR #268, opened 2020-08-04
Fix slug for sane

theodoreb:
Currently the word sane links to the "sanity-check" page making it difficult to view the definition for "sane"


theodoreb:
That seemed to do the trick, link is correct now :)

ovlb (approved):
Good catch! 🐚

PR #291, opened 2020-08-29
Add attribution to ableism excerpt

### LINKED ISSUE #289: Add attribution to excerpt in ableism definition
The first line of the body of the definition of ableism is followed by an attribution, but there is no attribution in the excerpt. Given that the excerpt is displayed in isolation, it is effectively claiming the original words of another author.

I believe that this specific situation needs to be addressed with some urgency since this excerpt is currently visible in search results, and to that end, I have made a branch and will soon open a PR to address it.

oakthielbar:
This is a PR to address the concerns of #289 by including attribution to the original author of the quotation used in the excerpt of the definition of ableism.

### DIFF
--- a/11ty/definitions/ableism.md
+++ b/11ty/definitions/ableism.md
@@ -2,7 +2,7 @@
 title: Ableism
 slug: ableism
 defined: true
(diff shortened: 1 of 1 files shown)


tatianamac (approved):
🚀 LGTM

PR #293, opened 2020-08-29
Add context of harm to excerpt for "barbaric"

### LINKED ISSUE #290: Excerpt for "barbaric" missing context on harm
I noticed when reviewing the current excerpts that they reproduce the first line of the definition. Is that the intended format for excerpts? Some of the definitions lead with a line that explains the purely grammatical and linguistic function of a term without any context, and I worry that a direct copy of just the first line could perpetuate the harmful use of some terms.

The current excerpt for "barbaric" is an example where the context of harm is missing. I believe that it should be updated to reflect the values and goals of this project.

oakthielbar:
This is a PR to address the concerns of #290 by adding context around the term's harm to the excerpt for the definition of "barbaric".

### DIFF
--- a/11ty/definitions/barbaric.md
+++ b/11ty/definitions/barbaric.md
@@ -5,7 +5,7 @@ flag:
   text: 'Neo-Colonial/Racist slur'
   level: 'avoid'
 defined: true
-excerpt: something which is obscenely cruel; primitive; unsophisticated.
+excerpt: Something which is obscenely cruel; primitive; unsophisticated. In modern day usage, you can notice this rhetoric being employed almost always for the purpose of demonizing "foreign" (more often than not from the global south) cultures.
 speech: adjective
 reading:
   - text: 'is the word barbarian a slur?'


11ty/definitions/barbaric.md
```diff
@@ -5,7 +5,7 @@ flag:
   text: 'Neo-Colonial/Racist slur'
   level: 'avoid'
 defined: true
-excerpt: something which is obscenely cruel; primitive; unsophisticated.
+excerpt: Something which is obscenely cruel; primitive; unsophisticated. In modern day usage, you can notice this rhetoric being employed almost always for the purpose of demonizing "foreign" (more often than not from the global south) cultures.
```
oakthielbar, line 8: The addition to this excerpt is pulled from the body of the definition. I left the line unmodified for the excerpt, but I am unsure if there are constraints on the length of the excerpt which would require changes.

tatianamac: AFAIK, it can vary, but it's around 320 in most instances, I believe.

tatianamac (approved):
🚀 LGTM!

PR #294, opened 2020-08-30
Include flag context in excerpt

### LINKED ISSUE #280: 🚩 Include flag context in excerpt
## Current Behaviour
The excerpt, which is used for social media and search engine previews, currently automatically prepends the word "avoid" if a term should be avoided and has an "avoid" flag in its definition.

## Desired Behaviour
The excerpt could also append the circumstances, such as "avoid: racist term."

## Notes
- See #262 for further details.

yourpalkat:
Closes #280 

Adds context for "avoid" flag in excerpt text used in social sharing, when possible. 

If text describing why a term should be avoided is included in frontmatter, the meta description will read (for example) "avoid: ableist language. definition excerpt" If "text" is not included under "flag", then the meta description will just read "avoid: definition excerpt".


ovlb:
Hey Derek,

Thanks so much for this. This fixes the underlying issue. 

I’ve one idea: by now the excerpt/meta description logic has grown into something inherently complex, I think I t makes sense to remove it from the template and use a [shortcode](https://www.11ty.dev/docs/shortcodes/) instead, as we would be able to write unit tests for it. 

Let me know if this is something you would be up to. If not – which is of course totally fine, as it would broaden the scope of the PR – I will create a follow-up issue. 

Thanks again!
O*

yourpalkat:
Thanks, Oscar – that makes sense! I'm happy to take a crack at it.

ovlb:
@derekmurr Great! If something trips you up, you can always ping me :)

yourpalkat:
@ovlb Ready for a review, I think, whenever you like! 

- Moved logic into a shortcode
- Added a little extra so the sentences are capitalized properly
- Wrote a couple tests, though I don't have much experience with them and may not have written them well...

ovlb (approved):
This looks great! Here’s the sea( shel)l of approval. 🐚

Thanks a ton, @derekmurr. Hope you had a pleasant experience and we’ll see you again soon. If anything stung, please let us know anytime :)

PR #315, opened 2020-09-01
🏳️ Remove Flag from Undefined Words

### LINKED ISSUE #312: 🐛 Define and reassess flags
See #310 and #311 . tl;dr is that the flags without context cause harm because people don't know what they mean and without the context of the definition they can cause harm.

**Action items:**
- Assess flags at global full-page view (remove for undefined terms)
- Rename "avoid" within codebase to "flag"
- Add definitions and context for all flags
- Link context for all flags.

ovlb:
As a step towards #312, this PR removes flags from all undefined words in the table of content.

This PR also adds tests to the ToC flag logic, to better document exceptions and desired outcomes, and remove fragility from the template files.


tatianamac:
Thank you, @ovlb ! Very helpful. As part of this PR, I think it makes sense to:

1. Remove all flags from the table of content level, and move the flag warnings down into part of the definition where it applies. That way, terms like OCD (which stemmed the curiosity around the flags), won't be flagged at the top tier. 

**Separate but related edits**, which I can open as a separate PR:
1.  Add flags at the lower, relevant levels within their definition. For example, add the medically appropriative aspect of a term in its own flagged section. Terms will then get their flags, but with added context, which I think is a win-win. I'm happy to provide more context and/or open another PR around *adding this* as it's a non-trivial lift.
2. Turn `hover` underline to white or black in the content warning box, as it currently fails contrast as red.
Content warning box with hover activated; background is mustard, underline is red

tatianamac (approved):

ovlb:
> Remove all flags from the table of content level, and move the flag warnings down into part of the definition where it applies. That way, terms like OCD (which stemmed the curiosity around the flags), won't be flagged at the top tier. 

If I read this correctly, the ToC should show no flags at all thereafter?

> Add flags at the lower, relevant levels within their definition. For example, add the medically appropriative aspect of a term in its own flagged section. Terms will then get their flags, but with added context, which I think is a win-win. I'm happy to provide more context and/or open another PR around adding this as it's a non-trivial lift.

Are you thinking about expanding the `flag` front-matter object with this needed context or integrating the flag via placeholders or sth similar to the definition content?

We can use this PR for this fix, too. Otherwise we’d need to remove the linked issue, I guess, as the PR fixes only a part of the needed steps.

> Turn hover underline to white or black in the content warning box, as it currently fails contrast as red.

Good catch. I’ll update this.

tatianamac:
> If I read this correctly, the ToC should show no flags at all thereafter?

Correct. I think that context-less, it causes more issues than it prevents. It also solves the issue of some terms having multiple flags. The visual indicator it causes can be overalarmist. 
 
> Are you thinking about expanding the `flag` front-matter object with this needed context or integrating the flag via placeholders or sth similar to the definition content?

It's a bit difficult to provide over text, but essentially, "redesigning" each of the sections to accommodate the flags. For example, "medical appropriation" would come with the general warning, and any additional insights. Think of it as a "If medical appropriation exists, include this module."

The content warning flag is the closest thing we have to this right now.

> We can use this PR for this fix, too. Otherwise we’d need to remove the linked issue, I guess, as the PR fixes only a part of the needed steps.
I think that we can link multiple issues and check off individual tasks manually, no? In general I'm more of a fan of smaller, more concise PRs. We can update the description of the original issue in order to accommodate?

ovlb:
@tatianamac Just found this age-old branch. I guess this is something we still want to do? I finally have some time again to allocate to Self-Defined and can finish this off.

Issue #271, opened 2020-08-22
🐛 Broken List Display in Southeast Asian Definition


ovlb:
The country list in the [definition of souteast asian](https://www.selfdefined.app/definitions/southeast-asian/) is forced into the first grid column, making it too small to read. We’d probably need to update the sibling selectors in the definition content CSS to move them to the second column or let them spread across two columns.

ajtran303:
I love CSS!

ajtran303:
Hi folx, I have a break from school so I am diving into this ticket during my short weekend!

ajtran303:
Here is a screenshot of what I came up with, what do you think? @ovlb 

<img width="1143" alt="Screen Shot 2020-09-07 at 5 18 23 PM" src="https://user-images.githubusercontent.com/31839316/92420053-567c5580-f12e-11ea-915d-064bbec3bdb7.png">

ajtran303:
Here is a screenshot of what I came up with, what do you think?  @tatianamac @ovlb

<img width="1143" alt="Screen Shot 2020-09-07 at 5 18 23 PM" src="https://user-images.githubusercontent.com/31839316/92420053-567c5580-f12e-11ea-915d-064bbec3bdb7.png">

ajtran303:
It looks like someone put in a PR for this too but have a different approach to solve in #321

PR #321, opened 2020-09-04
Fix mis-configured grid for <ul>s inside definitions

4shub:
## Summary
This PR was made to fix the following UI issue found [here](https://www.selfdefined.app/definitions/southeast-asian/)

### Before
Screen Shot 2020-09-04 at 4 50 55 PM

### After
Screen Shot 2020-09-04 at 4 50 44 PM

## Details
It looks like only paragraph level list items have the `grid-column` configuration css included. This adds the `grid-column` attribute to any `ul` item that's a direct child of the definition content. 

*note*: Not sure if this app might benefit if all "block" items should receive this `grid-column` treatment. But for now this change should fix the visual blip.

### DIFF
--- a/assets/css/structures/_definition-content.scss
+++ b/assets/css/structures/_definition-content.scss
@@ -35,6 +35,10 @@
       grid-column: 1 / -1;
       line-height: 1.4;
     }
+    
+    & > ul {
+      grid-column: 1 / -1;
+    }
 
     & h2 {
       flex: 0 1 auto;


ajtran303:
Hi! I made a branch to work on this too. Check out #271 -- I have a different approach to solve in #322

tatianamac:
Thank you, @4shub! I love how detailed your PR is. I think in this case the solution proposed in #271 fixes an additional problem of the countries list being less scannable (which was a prior issue, and not your fault). 

Re: this as a general list problem, I think fixing the inheritance as in #347 will solve this problem in a more global way. For now I'm going to mark this as duplicative. I really hope you'll submit again soon!

PR #328, opened 2020-09-08
Grammar fix on "Minoritised"

kittensyzygy:
Added punctuation and a determiner.


11ty/definitions/minoritised.md
```diff
@@ -12,7 +12,7 @@ reading:
     href: https://www.theodysseyonline.com/minority-vs-minoritize
 ---
 
-groups resulting from social constructs have been granted less power or representation compared to other members or groups in society
+groups, resulting from social constructs, that have been granted less power or representation compared to other members or groups in society
```
tatianamac, line 15: I'm not sure that "resulting from social constructs" is an apositive as I think the socially constructed part is critical.

I wonder if to a void the awkwardness of the original phrasing, we could instead say, "socially constructed groups that have been..." What do you think about that alternate approach?

tatianamac:
@kittensyzygy Are you still interested/do you have still capacity to work on this definition? No worries if not! I'm mostly hoping to get status updates on all the open PRs. Thank you for all you've done on it so far!

tatianamac (approved):

PR #346, opened 2020-10-07
Add a description to "agender"

erbridge:
This is based on my own experience as an agender person, and what I've learned from other agender people. It's almost certainly missing some nuance that other agender people have in their own meaning and use of the word.

First contribution, so let me know if I'm missing anything.


erbridge:
#347 fixes the layout issue in the preview.

11ty/definitions/agender.md
```diff
@@ -1,5 +1,16 @@
...
 title: agender
 slug: agender
-defined: false
----
\ No newline at end of file
+defined: true
+speech: adjective
+---
+
+An agender person is someone for whom one or more of the following is true:
```
tatianamac:
```suggestion
of, relating to, or characterised by being for whom one or more of the following is true:
```

11ty/definitions/agender.md
```diff
@@ -1,5 +1,16 @@
...
+
+- They are genderless, or without a gender
+- Their gender is neutral (neither a man nor a woman)
```
tatianamac:
```suggestion
- Their gender is neutral (neither a man nor a woman)
```
I'm curious here about reinforcing the gender binary with the parenthetical statement? I worry that this reinforces the idea that being genderneutral is somewhere "in between" if that makes sense. Though it's not explicitly defined, I worry about the implicit tendency here.

11ty/definitions/agender.md
```diff
@@ -1,5 +1,16 @@
...
+- Their gender is undefineable using the binary gender model
+- They don't know, understand, or care about their gender
+- They don't want to label their gender
+
+The term "non-binary" is often used to include agender people. While some agender people are also non-binary, not every agender person is comfortable being called non-binary or included in "non-binary" as an umbrella term.
```
tatianamac:
```suggestion
The term "non-binary" is often used to include agender people. While some agender people are also non-binary, not every agender person is comfortable being called non-binary or included in "non-binary" as an umbrella term.
```

To put a finer point on this part, I think we should break out this into two separate ideas, perhaps under a "Note" sections (a good example of this type of construction is here, under (Hispanic)[The term "non-binary" is often used to include agender people.])
### Notes
1. Some agender people also identify as non-binary, gender non-conforming, genderfluid, and/or gender queer, but being agender does not preclude being any of those things.
2. Cisgender and transgender people can be agender. Being agender does not preclude either of those things.
3. The term "non-binary" is often used to include agender people, when as mentioned above, some agender people identify as both while others do not. Be mindful of not making this assumption when using the term.

tatianamac (changes requested):
Hi @erbridge ! I keep making comments and not submitting the full review! Here are some notes!

11ty/definitions/agender.md
erbridge: I think for many readers the statement without the parenthetical is really hard to understand without a point of reference in binary genders. Having said that, I'm very open (eager?) for a way of expressing this concept in an accessible way without needing to refer to binary genders.

erbridge: Thinking more, I think it's probably fine to drop the parenthetical, so I have.

tatianamac (approved):
🚀  Looks good to me!

tatianamac:
Thank you for your first contribution, @erbridge ! I hope you'll return again as I appreciate your thoughtfulness and you supporting our project!

PR #347, opened 2020-10-07
Style `ul`s like `p`s when they're direct content

erbridge:
Without this change, `ul`s get compressed to a single grid column if they're direct children of `.definition-content__content`.

# Before

image

# After

image


tatianamac (approved):
🚀

tatianamac:
Thank you, @erbridge ! 🙏🏽

PR #368, opened 2020-11-01
<<WIP>> Define the n-word

ceeoreo:
I included a definition for the n-word. I would like to include the Issue section along with Impact. I've noticed that many other definitions include outside reference material as well. Are there any suggestions or guidelines as far as references go? I want to include credible sources and I would prefer they be majority historical opposed to think-pieces. Opinions on this topic can be highly offensive since this term is a very sensitive area for many. (also this is my first time contributing to oss so i hope i'm doing this correctly 😅)

### DIFF
--- a/11ty/definitions/n-word.md
+++ b/11ty/definitions/n-word.md
@@ -1,8 +1,13 @@
 ---
 title: N-word
 slug: n-word
-defined: false
+defined: true
 flag:
   level: avoid
   text: racist slur
 ---
+The n-word is a euphemism for the term "nigger" or "nigga". It is considered _highly_ offensive for non-Black people to refer to anyone, especially Black people, as the "n-word", no matter the variant of the word. It has been reclaimed by Black people and is often used amongst themselves.
+
+## Issues
+
+The "n-word" was used to refer to enslaved Black people and later became normalized as a derogatory way to refer to Black people. The "n-word" was used in literature and other media. However, by the 1960s, the use of this term by non-Black people became widely recognized as offensive and inappropriate.


tatianamac:
@ceoraford ! Thank you for submitting this. You did great! 🎉 

**To answer your questions:**
- **Re: sources,** we tend to try to pick verified sources like newspapers, journals, etc. AND we try to balance that with first-person accounts (like blogs, personal articles, etc.). It's important to me that we share "real" voices, and don't only validate mainstream sources. I often reference NPR Codeswitch and other pubs that focus on the nuance of language.
- For a complex term like this, I think it would be great to add nuance and multiple perspectives. Would you feel comfortable asking Black folks in Self-Defined Slack who are most impacted by this word to contribute their perspectives?
- For a reference to a more nuanced term, I think [pronouns](https://www.selfdefined.app/definitions/pronouns/) could be helpful.

**Some questions I have, fully recognising I am a non-Black person:**
- We wrote n-word meaning the full word here, much like we did with [r-word](https://www.selfdefined.app/definitions/r-word/). I'm wondering if adapting a structure similar to that might help.
- I'm curious if adding a bit of the historical context and power/connection to white supremacy and slavery would help in a "Context" section.
- I am curious about adding something about non-Black people not regulating the usage of the n-word among Black folks.
- I am also curious about adding the idea of "taking back" terms, and that this is an opinion that many but not all Black folks have (and thus, reinforcing the idea that Black folks aren't a monolith and won't agree on this).
- It might be good in that above example to note that even with take-back terms, it can still cause harm to others who find the term hurtful and offensive. So, it's a choice.

Hopefully that helps! Let me know if I can support you in any other way. I know a term like this requires a lot of emotional labour, so I am very appreciative of you spending it here! Once we get more community feedback, I'll provide a more editorial overview in my PR review (looking for consistency, syntax, etc.)

PR #369, opened 2020-11-01
<<WIP>> Definition/blindly

ceeoreo:
I have a basic idea of a definition for blindly. I was thinking something like "Term used to characterize being unaware or oblivious. "Blindly" is considered ableist language and should be avoided". Wanted a second opinion before I added this though

### DIFF
--- a/11ty/definitions/blindly.md
+++ b/11ty/definitions/blindly.md
@@ -1,0 +1,8 @@
+---
+title: Blindly
+slug: blindly
+defined: false
+flag:
+  level: avoid
+  text: ableist language
+---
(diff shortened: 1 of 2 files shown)


tatianamac:
@ceoraford Absolutely! A few ideas/thoughts I have to add:

- I think it would be awesome if we cited some resources in the definition, particularly those from Blind folks.
- [Autistic Hoya](https://www.autistichoya.com/p/ableist-words-and-terms-to-avoid.html) has a good start for blindly and why it can be ableist.
- We could add some examples like "Blind leading the Blind"

Let me know if you have any other thoughts or if I can support you and this PR in any other way!

PR #373, opened 2020-11-30
Rename "colourblind" definition file

ashur:
## Description

Currently, the link from the main page to the **colourblind** article is broken. On closer inspection, the definition filename uses the `color` spelling variation, while the `slug` uses `colour`.

## Changes

This PR makes one change:

- Rename `11ty/definitions/colorblind.md` → `11ty/definitions/colourblind.md`

## Notes

The definition file's frontmatter uses the `u` variation of `colour` twice:

```yaml
title: colourblind
slug: colourblind
```

leading me to believe that is the intended spelling. Rather than change those values to match the file name, I renamed the file to match the frontmatter.

## Personal Notes

I'm very sorry if this is the wrong way to go about fixing a broken link in this case where I'm not 100% positive about the best path forward! I didn't see anything related in the contributing guidelines, but if it would be more appropriate to open an issue I'm more than happy to do that.

Thanks!


tatianamac (approved):
🚀 Thank you!

tatianamac:
This is perfect. Thank you for catching that error and opening up a PR, @ashur! I'll open an issue to make renaming files more clear (I also owe a "style guide" type post to the docs to clarify that we use International English. 🙃)

PR #388, opened 2021-03-30
Fix typo in oriental.md

unwitting:
Removes the word "present" which I think looks to be a typo?


tatianamac (approved):
Great catch. Thank you!

PR #397, opened 2021-04-27
#392 Prevent double-escaping, same as `content`

### LINKED ISSUE #392: Ampersand character rendered as raw HTML entity code
On the _antisemitism_ definition page in the "_Further Reading_" section there are some links, one of which has an _ampersand_ character that is being rendered as an HTML entity code instead of the logogram that would be expected in an anchor text.

qrazi:


qrazi:
The lighthouse check seems to fail, but I'm not sure why. I mean, I get the error, but I don't see how my code changes cause lighthouse to fail? https://storage.googleapis.com/lighthouse-infrastructure.appspot.com/reports/1619523007710-75919.report.html

tatianamac:
@qrazi That was my bad! I had merged in a PR to add lighthouse, which had an outstanding bug. I resolved the bug, so this should pass and then I can merge it in! 🤞🏽

tatianamac (approved):
🚀  Thank you for this fix, @qrazi !

PR #398, opened 2021-04-27
⚒️ Resolves lighthouse issue

tatianamac:
## Problem
We installed Lighthouse as part of the checks, but it's causing a tap target failure (which has nothing to do with folks' PRs). 

## Fix
Changed list styling for Further Reading section, making it more legible in the process, which will also prevent downstream issues.


mikerapin (approved):

tatianamac:
Thank you, @mikerapin !

PR #400, opened 2021-04-28
fix(ui): hide "Ways to help" section under a spoiler

### LINKED ISSUE #325: Relocate/adapt "Ways to help" section on mobile view
On mobiles, it takes a number of scrolls in the UI before reaching the delicious dictionary definition content. Could the "Ways to help" section be adapted to allow the user to reach the definitions quicker and place the "Ways to help" (which might be relevant to a more niche number of the site users), e.g., at the bottom of the screen (with perhaps some quick links in the current location in the header) or inside an accordian, etc.?

embarq:
Closes https://github.com/selfdefined/web-app/issues/325


embarq:
It was made to hide the "Ways to help" content only on mobile. I found that for the parent section `800px` breakpoint is used, so I took the same value into consideration.

Currently, the script will collapse the section on <800px screens, for desktop it will look as it was before.

However, the spoiler behaviour won't change on resize. For example if you open the homepage on mobile-like screen, and then resize it to let's say 810px - the spoiler will remain working(perhaps it should not). If you would like to have it adjusting to resize, let me know.

11ty/index.njk
```diff
@@ -15,9 +15,24 @@ layout: layouts/index.njk
...
+    <h2 class="sub-headline">Ways to help
+      <button
+        aria-expanded="true"
+        aria-controls="ways-to-help-content"
+        class="spoiler__toggle-btn"
+        id="ways-to-help-visibility-toggle">
+        {# https://css.gg/chevron-down #}
+        <svg xmlns="http://www.w3.org/2000/svg" width="24" height="24" fill="none" viewBox="0 0 24 24"><path fill="currentColor" d="M6.34 7.76L4.93 9.17 12 16.24l7.07-7.07-1.41-1.41L12 13.4 6.34 7.76z"/></svg>
+      </button>
+    </h2>
```
ovlb: Blocking: The button has no accessible name at the moment ([see this article for reference](https://www.sarasoueidan.com/blog/accessible-icon-buttons/)). We need to add some to have it make sense for screen reader users.

assets/css/layouts/_homepage.scss
```diff
@@ -4,4 +4,16 @@
...
+  .sub-headline {
+    align-items: center;
+    display: flex;
+    justify-content: space-between;
+
+    .spoiler__toggle-btn {
+      color: #fff;
+    }
+  }
+}
```
ovlb: I wonder if we can make these style more reusable, and less tied to the specific instance?

11ty/index.njk
```diff
@@ -15,9 +15,24 @@ layout: layouts/index.njk
...
+    <ol
+      class="list spoiler__content"
+      id="ways-to-help-content"
+      aria-labelledby="ways-to-help-visibility-toggle">
```
ovlb: Blocking: This `aria-labelledby` does not work, as it’s pointing to the empty button. If we add toggle text to the button, this is probably not what we want as a label for the list. I’d propose to point to the headline, and move the button out of the headline (e.g. having the button and headline in a `header`. 

```suggestion
      aria-labelledby="ways-to-help"
```

ovlb (changes requested):
Hey! Thanks so much. This will come in handy. I’ve left some accessibility-related notes. :)

assets/css/layouts/_homepage.scss
embarq: Do you mean both `.sub-headline` and `.spoiler__toggle-btn`?

ovlb: Yes, the combination of both. Aligned with the proposed update of the structure (https://github.com/selfdefined/web-app/pull/400#discussion_r621928910) you should be able to remove the nested styles altogether. Maybe add a utility class `.flex-space-between` to have it reusable and apply this to the header. Am I making sense? Have a bit of a brain freeze currently.

ovlb: Another thing, for line 16. We added a dark mode last year, and as such, we try to avoid hardcoded colour values. The [links in the box are styled in this file](https://github.com/selfdefined/web-app/blob/prod/assets/css/components/_box.scss#L17), maybe it works to apply the same styles to the button?

11ty/index.njk
embarq: The article was quite helpful, thanks a lot for sharing!

embarq:
@ovlb Man, I can't express how grateful I am. Thanks for so many useful and interesting suggestions!
I tried my best in an attempt to make the implementation better, would be awesome if you could check the result once again.

P.S. Thanks for being so helpful 👍 As you might notice, I have quite a limited knowledge of a11y and how to implement it in the right way. Looking forward to learning and practice more tho

11ty/index.njk
embarq: Fixed in c73385a

11ty/index.njk
embarq: Fixed in c73385a

assets/css/layouts/_homepage.scss
embarq: As for utility classes, here are related commits: 641120c, ff605f6
As for wrong icon color, I choose accent color. I think it looks even cooler 😁 . 8b69cce

ovlb:
@embarq Thanks for your updates. I had quite a packed week, so please bear with me. I want to review it with the time it deserves. Aiming for Monday. Hope you have a nice weekend. 

O*

PR #404, opened 2021-04-30
Update to node 14, update dependencies

mpcsh:
I've been noticing some `npm` warnings; this PR addresses them.
- `node` updated from v12 to v14 (latest is 14.16.1 as of now)
- Ran `npm audit fix` and `npm update` (these do not cross semver major boundaries)

Things that'd be nice to do:
- Address `npm outdated` output:
```
❯ npm outdated
Package                 Current  Wanted  Latest  Location
autoprefixer              9.8.6   9.8.6  10.2.5  selfdefined
concurrently              5.3.0   5.3.0   6.0.2  selfdefined
eslint                    6.8.0   6.8.0  7.25.0  selfdefined
eslint-config-prettier   6.15.0  6.15.0   8.3.0  selfdefined
husky                     4.3.8   4.3.8   6.0.0  selfdefined
markdown-it-anchor        5.3.0   5.3.0   7.1.0  selfdefined
markdownlint-cli         0.23.2  0.23.2  0.27.1  selfdefined
prettier                 1.19.1  1.19.1   2.2.1  selfdefined
```
- Address warnings on `npm run serve`:
```
[1] DEPRECATION WARNING: $hue: Passing a unit other than deg (120s) is deprecated.
[1] 
[1] To preserve current behavior: $hue / 1s
[1] 
[1] See https://sass-lang.com/d/color-units
[1] 
[1]   ╷
[1] 7 │   dark-green: hsl(120s, 100%, 15%),
[1]   │               ^^^^^^^^^^^^^^^^^^^^
[1]   ╵
[1]     assets/css/abstracts/_variables.scss 7:15  @import
[1]     assets/css/base.scss 4:9                   root stylesheet
[1] 
[1] WARNING: You probably don't mean to use the color value red in interpolation here.
[1] It may end up represented as red, which will likely produce invalid CSS.
[1] Always quote color names when using them as strings or map keys (for example, "red").
[1] If you really want to use the color value here, use '"" + $name'.
[1] 
[1]   ╷
[1] 4 │     --clr-#{$name}: #{$value};
[1]   │             ^^^^^
[1]   ╵
[1]     assets/css/base/_custom-properties.scss 4:13  raw-color-names()
[1]     assets/css/base/_custom-properties.scss 51:3  @import
[1]     assets/css/base.scss 12:9                     root stylesheet
```
⚠️ _Note: this output is incomplete and abridged._

PR #410, opened 2021-05-09
Removes .nvmrc; gitignores .nvmrc and .tool-versions

mpcsh:
See #404 for context, specifically [this comment](https://github.com/selfdefined/web-app/pull/404#issuecomment-834756503) and [this comment](https://github.com/selfdefined/web-app/pull/404#issuecomment-835721761)


ovlb:
Hey. We should not remove these files, as they configure different tools to use the correct NPM version. To avoid errors in the build we should make sure that all contributors use the same NPM version not configure them as they see fit.

ovlb:
For context: we had some issues with folx using NPM versions that were all over the place and dependencies that failed in these versions. Since we added the tool files to have canonical versioning.

ovlb (changes requested):
See comments.

mpcsh:
Fair enough! See #412

PR #435, opened 2021-09-01
✏️ Edits 'dumb' to add related words, more clarity

### LINKED ISSUE #365: Define "dumbfounded" & "dumbstruck"
Both of these build on the ableism inherent in "[dumb](https://www.selfdefined.app/definitions/dumb/)"
We could make these definitions children of that word.

Some alternatives:
astounded
awestruck
flabbergasted
nonplussed

Some quick research:
https://cishits.tumblr.com/post/79281810833/two-things-1-i-feel-like-ableist-language-is
https://www.quora.com/Do-you-know-when-you-are-using-derogatory-ableist-language/answer/Jayden-Brunner

tatianamac:
#365

## Overview
- Adds synonyms per #365 with additional resources
- Cleans up some of the language
- Adds additional suggested words

### DIFF
--- a/11ty/definitions/dumb.md
+++ b/11ty/definitions/dumb.md
@@ -17,11 +17,11 @@ alt_words:
     - ridiculous
     - shallow
     - silly
-    - senseless
     - strange
     - weird
     - unfamiliar
     - unfathomable
+    - unreasonable
     - unnecessary
   dumbfounded (noun):
     - amazed
--- a/11ty/definitions/dumb.md
+++ b/11ty/definitions/dumb.md
@@ -8,18 +8,30 @@ defined: true
 excerpt: non-verbal, non-speaking, silent, refraining from speaking or unable to speak; sometimes __mute__ which can also be ableist; colloquially used to mean unintelligent or senseless
 speech: adjective
 alt_words:
-  - basic
-  - incomprehensible
-  - nonsensical
-  - redundant
-  - shallow
-  - silly
-  - senseless
-  - strange
-  - weird
-  - unfamiliar
-  - unfathomable
-  - unnecessary
+  dumb (noun):
+    - basic
+    - incomprehensible
+    - nonsensical
+    - redundant
+    - shallow
+    - silly
+    - senseless
+    - strange
+    - weird
+    - unfamiliar
+    - unfathomable
+    - unnecessary
+  dumbfounded (noun):
+    - amazed
+    - awed
+    - floored
+    - nonplussed
+  dumbstruck (noun):
+    - astonished
+    - perplexed
+    - shocked
+    - stunned
+    - stupified
 reading:
   - text: Etymonline entry
     href: https://www.etymonline.com/word/dumb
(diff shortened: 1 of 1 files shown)


11ty/definitions/dumb.md
```diff
@@ -8,18 +8,30 @@ defined: true
...
+    - amazed
+    - awed
+    - floored
+    - nonplussed
+  dumbstruck (noun):
+    - astonished
+    - perplexed
+    - shocked
+    - stunned
+    - stupified
```
Primigenus: I think stupefied carries the same ableist loaded meaning as dumbfounded, right?

11ty/definitions/dumb.md
```diff
@@ -8,18 +8,30 @@ defined: true
...
-  - senseless
-  - strange
-  - weird
-  - unfamiliar
-  - unfathomable
-  - unnecessary
+  dumb (noun):
+    - basic
+    - incomprehensible
+    - nonsensical
```
Primigenus: Should we avoid recommending any language connecting senses or intelligence to lack of knowledge?

11ty/definitions/dumb.md
```diff
@@ -8,18 +8,30 @@ defined: true
...
+    - redundant
+    - shallow
+    - silly
+    - senseless
```
Primigenus: Same as above comment

Primigenus (commented):
Love this, thank you, just some minor questions around some alt suggestions. Apologies for not catching your request back in November!

11ty/definitions/dumb.md
tatianamac: Good catch!

11ty/definitions/dumb.md
tatianamac: I get what you're saying—my understanding is nonsense is connotes sense in the meaning of reason. From what I've read particularly from Deaf folks, it seems like there's a desire to not overcorrect (i.e., avoid words like "I see what you mean". With that said, I'd rather err on the side of caution so I can remove it, but wanted to provide the context!

tatianamac:
@Primigenus Thank you for the thoughtful notes, Rahul! I pushed up some changes; let me know what you think—if good, will you approve? Thank you!

(Also no sorries!)

11ty/definitions/dumb.md
Primigenus: Yeah, I've heard similar things from Deaf folks. It feels like maybe there's an heuristic here where we avoid using it as an adjective but are OK with it as a noun? Eg. "that's nonsense" seems alright to me, but "you're being nonsensical" implies that potentially ableist comparison. What do you think?

Primigenus (approved):
LGTM, thank you Tatiana!

11ty/definitions/dumb.md
tatianamac: I can see that! Either way I agree we should remove!

Primigenus (approved):

PR #436, opened 2021-09-02
➕ Adds spongebob case

### LINKED ISSUE #417: Spongebob case
I'm not seeing a good way to add information about the alternating upper and lower letters, which comes from mocking how certain disabled people speak (before that from "a meme where Spongebob is bent at odd angles and often has that text superimposed in that casing on it" - Tatiana, hence the ticket title).

Ie, is there a good way to make entries for a more general pattern, rather than for a specific word?
I'm assuming this could start with a "spongebob case" entry, just that I'd be surprised if I'm alone in not previously knowing that's what it's called, so probably wouldn't have found it.

tatianamac:
Closes #417.

Adds draft definition for SpongeBob case.

### DIFF
--- a/11ty/definitions/spongebob-case.md
+++ b/11ty/definitions/spongebob-case.md
@@ -1,0 +1,38 @@
+---
+title: SpongeBob case
+slug: spongebob-case
+flag:
+  text: 'ableist language'
+  level: avoid
+defined: true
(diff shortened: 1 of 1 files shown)


FennelAurora (approved):
Thanks @tatianamac (and sorry for the delay)! I'm not someone who is targetted by this form of abuse, so my approval has limited value here - still for me this is a very clear and good entry, so approved.

PR #446, opened 2021-11-13
fix: Prevent flash of light mode if dark mode is selected

eps1lon:
Instead of initializing the theme in a deferred script we now initialize it before any html is parsed. This prevents a flash of light mode when light mode is selected between the page being rendered and JS being executed.

### DIFF
--- a/11ty/_includes/layouts/base.njk
+++ b/11ty/_includes/layouts/base.njk
@@ -35,6 +35,29 @@
     {% endblock pageStyles %}
   </head>
   <body>
+    <script>
+      /**
+      * @type String
+      */
+      let userThemeSetting = localStorage.getItem('sdUserTheme');
+
+      if (userThemeSetting) {
+        document.documentElement.setAttribute(
+          'data-user-theme',
+          userThemeSetting
+        );
+      }
+
+      window.setUserPreference = function setUserPreference(value) {
+        localStorage.setItem('sdUserTheme', value);
+        document.documentElement.setAttribute('data-user-theme', value);
+      }
+
+      window.unsetUserPreference = function unsetUserPreference() {
+        localStorage.removeItem('sdUserTheme');
+        document.documentElement.removeAttribute('data-user-theme');
+      }
+    </script>
     {% block content %}{% endblock content %}
     {% include 'components/base/site-footer.njk' %}
     {% block pageScript %}
--- a/11ty/assets/js/theme-switcher.js
+++ b/11ty/assets/js/theme-switcher.js
@@ -9,17 +9,15 @@ function initThemeSwitch() {
 
   let $buttons = $switchContainer.querySelectorAll('[type="radio"]');
 
-  /**
-   * @type String
-   */
-  let userThemeSetting = localStorage.getItem('sdUserTheme');
-
   function setInitialState() {
+    /**
+     * @type String
+     */
+    let userThemeSetting = document.documentElement.getAttribute(
+      'data-user-theme'
+    );
+
     if (userThemeSetting) {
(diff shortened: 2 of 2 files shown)


11ty/_includes/layouts/base.njk
```diff
@@ -35,6 +35,29 @@
...
+          userThemeSetting
+        );
+      }
+
+      window.setUserPreference = function setUserPreference(value) {
+        localStorage.setItem('sdUserTheme', value);
+        document.documentElement.setAttribute('data-user-theme', value);
+      }
+
+      window.unsetUserPreference = function unsetUserPreference() {
```
eps1lon, line 56: `window.unsetUserPreference = ` is redundant but I felt like it's safer to be explicit here.

11ty/_includes/layouts/base.njk
```diff
@@ -35,6 +35,29 @@
...
   <body>
+    <script>
+      /**
+      * @type String
+      */
+      let userThemeSetting = localStorage.getItem('sdUserTheme');
+
+      if (userThemeSetting) {
+        document.documentElement.setAttribute(
+          'data-user-theme',
```
malchata, line 46: Not sure if this was an intentional decision, but consider using [the `dataset` API](https://developer.mozilla.org/docs/Web/API/HTMLElement/dataset) for `data-` attributes on this and any other instances if possible.

tatianamac:
Congrats and thank you for your first PR, @eps1lon ! Appreciate you looking out for this sort of thing. 

I want to provide a bit of context around why we deferred initially: Rather than load the script unnecessarily for all, we wanted to only load as needed (which trade off, as you identified, is flash).

I am going to ask a few folks with more niche performance knowledge in this to provide more prescriptive feedback to your PR, including @malchata !  stay tuned for feedback. Thank you!

malchata (approved):
LGTM!

11ty/_includes/layouts/base.njk
eps1lon: This code just moved. I didn't change any implementation details beyond timing.

ovlb: I think this was me, so feel free to update this, @eps1lon, but it’s on me to do it, if you don’t want to. :)

ovlb:
Also fine with this. Is there an acronym for flash of unstyled light mode already?

🦭 of aprroval

ovlb (approved):

11ty/_includes/layouts/base.njk
malchata: Gotcha, no worries, then. No need to break things if they're working fine!

11ty/_includes/layouts/base.njk
```diff
@@ -35,6 +35,29 @@
...
+          userThemeSetting
+        );
+      }
+
+      window.setUserPreference = function setUserPreference(value) {
```
mpcsh, line 51: (Naming nitpick, non-blocking): while we're here, maybe this would be better named `setUserTheme` instead of `setUserPreference`? I say that because we're hardcoding `sdUserTheme` as the localstorage key down below, but to me the name `setUserPreference` implies that we can use this function to set arbitrary preferences (i.e. I'd expect it to take both a key and a value).

11ty/_includes/layouts/base.njk
mpcsh: As above, I wonder if this would be better as `unsetUserTheme`. Agree with the explicit unsetter method though!

mpcsh (approved):
Looking great!

